PR #6, opened 2025-03-28
Add IDAPython REPL Functionality

rand-tech:
This PR adds a new feature that enables running IDAPython code directly through the MCP and captures the output (stdout/stderr). 

NOTE: It will execute the code using `exec` so it is potentially vulnerable to prompt injection.

### DIFF
--- a/mcp-plugin.py
+++ b/mcp-plugin.py
@@ -228,6 +228,8 @@ import logging
 import queue
 import traceback
 import functools
+import sys
+import io
 
 import ida_hexrays
 import ida_kernwin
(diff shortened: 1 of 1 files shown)


mrexodia:
Thanks for your contribution! Was the function actually (naturally) used by the LLM in one of your use cases? I'm a bit hesitant to add functionality like this because it might hide symptoms of a missing tool.

h5kk:
> Thanks for your contribution! Was the function actually (naturally) used by the LLM in one of your use cases? I'm a bit hesitant to add functionality like this because it might hide symptoms of a missing tool.

Could perhaps specify to only use the other tool if no other tools are more relevant -- maybe that's what happens if the tool is at the bottom? Either way I think it would be great to have.

mrexodia:
For me the issue with this is that LLMs are super bad at writing IDAPython code (because the API changed so much over the years), so unless there is an actual example where the LLM decided to use this I don’t want to merge it.

While experimenting I noticed the quality of the tools is really important, so that’s why I’m so hesitant…

mrexodia:
It looks like similar functionality got exploited already, so better to not merge it for now 😬 https://x.com/junr0n/status/1905978316043219198

PR #10, opened 2025-03-29
make llm could define a struct in IDA

MrRtcl:


mcp-plugin.py
```diff
@@ -723,6 +723,26 @@ def set_comment(
             return
         cfunc.del_orphan_cmts()
         cfunc.save_user_cmts()
+    raise IDAError(f"Failed to set comment at {address}")
```
mrexodia, line 726: This check was removed on purpose. The set_comment will set on the disassembly first and then do a best effort for the decompilation.

mcp-plugin.py
```diff
@@ -723,6 +723,26 @@ def set_comment(
...
+    type: Annotated[str, "structure declaration in C syntax. Example: 'struct MyStruct { int x; int y; }'"]
+):
+    ret = idaapi.parse_decls(None, type, None, idaapi.PT_TYP)
+    if ret != 0:
+        raise IDAError(f"Failed to parse structure declaration: {type}")
+    
+
+@jsonrpc
+@idawrite
+def set_disassembly_comment(
```
mrexodia, line 740: This also has to be removed

mcp-plugin.py
```diff
@@ -723,6 +723,26 @@ def set_comment(
...
         cfunc.del_orphan_cmts()
         cfunc.save_user_cmts()
+    raise IDAError(f"Failed to set comment at {address}")
+
+@jsonrpc
+@idawrite
+def add_structure(
+    type: Annotated[str, "structure declaration in C syntax. Example: 'struct MyStruct { int x; int y; }'"]
+):
+    ret = idaapi.parse_decls(None, type, None, idaapi.PT_TYP)
```
mrexodia, line 733: Did you confirm this actually adds the structure to the local types view? I tried a few similar things, but it didn’t work…

mrexodia (changes requested):
Overall good, but needs some changes

mrexodia:
Thanks for your initial work on this. The code will be replaced with something different today.

mcp-plugin.py
MrRtcl: emm I think it's ok.

MrRtcl: image
I think it's ok ....
image

PR #11, opened 2025-03-31
Add string list and precise search

dDostalker:
First of all， I'm happy to see such a great project that makes MCP very easy for everyone to implement！

## Add function

- String exact search
- Lists the strings currently in ida

## Potential problems exist

- Should I change the return value of my new additions from List[str] to List[* specific type *]?
- When the target program is too large, the function of listing strings may not achieve the desired effect, and the precise search may lead to the omission of some keywords, should we add a fuzzy search function?

If there is anything that needs to be fixed, please let me know and I will be happy to improve it


dDostalker:
Now that it have a better implementation at #12 I think I can close this pr

PR #12, opened 2025-03-31
Implement list_strings and search_strings

zouxianyu:
Implement two functions about strings:
1. `list_strings`: List all strings.
2. `search_strings`: Search for strings containing the given pattern.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ Available functionality:
 - `convert_number(text, size)`: Convert a number (decimal, hexadecimal) to different representations.
 - `list_functions(offset, count)`: List all functions in the database (paginated).
 - `list_strings(offset, count)`: List all strings in the database (paginated).
-- `search_strings(pattern, offset, count)`: Search for strings containing the given pattern (case-insensitive).
+- `search_strings(pattern, offset, count)`: Search for strings that satisfy a regular expression.
 - `decompile_function(address)`: Decompile a function at the given address.
 - `disassemble_function(start_address)`: Get assembly code (address: instruction; comment) for a function.
 - `get_xrefs_to(address)`: Get all cross references to the given address.
--- a/README.md
+++ b/README.md
@@ -16,6 +16,8 @@ Available functionality:
 - `get_current_function()`: Get the function currently selected by the user.
 - `convert_number(text, size)`: Convert a number (decimal, hexadecimal) to different representations.
 - `list_functions(offset, count)`: List all functions in the database (paginated).
+- `list_strings(offset, count)`: List all strings in the database (paginated).
+- `search_strings(pattern, offset, count)`: Search for strings containing the given pattern (case-insensitive).
 - `decompile_function(address)`: Decompile a function at the given address.
 - `disassemble_function(start_address)`: Get assembly code (address: instruction; comment) for a function.
 - `get_xrefs_to(address)`: Get all cross references to the given address.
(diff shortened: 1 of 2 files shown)


dDostalker:
Very perfect implementation, your implementation is much better than mine!But I think it would be better to split the search into fuzzy search and precise search. Would you be interested in discussing that with me?

zouxianyu:
Sorry I missed your pull request while working on implementing the functions. I only noticed you'd already implemented the same feature after I finished. I'm submitting this pull request anyway to offer an alternative implementation.

You make a good point, adding regex support for search might be the better approach. What do you think?

dDostalker:
it doesn't matter, this is a good method, but I am worried that ai can not directly generate expressions, it may let the user give more hints, which is what I have been worried about, I try to write a regular expression function test. Another interesting thing is that we seem to be from the same country, and we can directly communicate with each other about these small details through QQ or WeChat

mrexodia:
I’m okay with merging this. Do you have a chat log where Claude used this tool successfully to achieve something?

dDostalker:
Of course, here are some of the chat logs I tested, and it worked even better than I expected.

1.
> user: 
> Search the current ida page for net-related keywords with more than 10 letters
> 
> assistant:
> The search for "net"-related keywords returned the following strings, but none of them are longer than 10 letters:
> 
> 1. **Address:** `0x407348`
>    - **String:** `\Microsoft\Internet Explorer\Quick Launch`
>    - **Length:** 42
> 
> 2. **Address:** `0x409038`
>    - **String:** 
>      ```
>      Installer integrity check has failed. Common causes include
>      incomplete download and damaged media. Contact the
>      installer's author to obtain a new copy.
> 
>      More information at:
>      http://nsis.sf.net/NSIS_Error
>      ```
>    - **Length:** 204
> 
> Would you like me to refine the search or look for something else?
2.

> user: 
> Search for strings that contain more than 2 special matches
> 
> assistant:
> Here are some strings from the database that contain more than 2 special characters (e.g., `\\`, `%`, `*`, `?`, `|`, `<`, `>`, `/`, `"`, etc.):
> 
> 1. **Address:** `0x4072bc`  
>    **String:** `.DEFAULT\\Control Panel\\International`
> 
> 2. **Address:** `0x4072e4`  
>    **String:** `Control Panel\\Desktop\\ResourceLocale`
> 
> 3. **Address:** `0x40731c`  
>    **String:** `Software\\\\Microsoft\\\\Windows\\\\CurrentVersion`
> 
> 4. **Address:** `0x407348`  
>    **String:** `\\\\Microsoft\\\\Internet Explorer\\\\Quick Launch`
> 
> 5. **Address:** `0x409038`  
>    **String:** `Installer integrity check has failed. Common causes include\nincomplete download and damaged media. Contact the\ninstaller's author to obtain a new copy.\n\nMore information at:\nhttp://nsis.sf.net/NSIS_Error`
> 
> 6. **Address:** `0x4091cc`  
>    **String:** `%u.%u%s%s`
> 
> 7. **Address:** `0x409334`  
>    **String:** `%s=%s\r\n`
> 
> 8. **Address:** `0x40933c`  
>    **String:** `*?|<>/\\":`
> 
> If you'd like to explore any of these strings further or analyze their context, let me know!

mrexodia:
Looks great, thank you for the contribution @dDostalker and @zouxianyu

PR #13, opened 2025-03-31
Fixed `SyntaxWaring: "is not" with a literal`

winl0gon:
Just simple SyntaxWarning fixed
 
```
$HOME/.idapro/plugins/mcp-plugin.py:837: SyntaxWarning: "is not" with a literal. Did you mean "!="?
  if result is not 0:
```

Ref.
- https://adamj.eu/tech/2020/01/21/why-does-python-3-8-syntaxwarning-for-is-literal/


mrexodia:
Thanks, but I’m about to replace this code with something completely different.

PR #14, opened 2025-04-01
add rename & retype for gloabl variable

MrRtcl:


mrexodia:
Thanks for your contribution!

PR #17, opened 2025-04-03
Add the highlight functionality to disassembly code

dDostalker:
At first I tried to add this functionality to both decompile and disassemble code，but it doesn't run correctly in the decompile code . So I had to remove that part of the functionality. I think I need more time to finish the `highlight_decompile`.

Below is a demo of the functionality:

https://github.com/user-attachments/assets/c7ec3e4f-6a33-4fd0-a50c-c39023775612

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,7 @@ Available functionality:
 - `set_function_prototype(function_address, prototype)`: Set a function's prototype.
 - `declare_c_type(c_declaration)`: Create or update a local type from a C declaration.
 - `set_local_variable_type(function_address, variable_name, new_type)`: Set a local variable's type.
+- `highlight_disassemble` : Highlight disassembly lines or functions
 
 ## Prerequisites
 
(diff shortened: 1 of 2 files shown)


mrexodia:
image

In the video it also seems to highlight the `add` instruction, even though you told it to highlight the `mov` instruction?

While the functionality itself seems to work mostly fine, what would be the benefit to the LLM workflow? There is no additional information that the LLM gets later from highlighting instructions and it will just add more possibility of tool confusion in my view. If we really wanted to implement something like this we could prompt the LLM to start comments with `IMPORTANT` and then transparently trigger the highlighting functionality when that keyword is found/removed. This way there is no additional tool that the LLM can get potentially confused about and you would get the same result.

dDostalker:
Well, originally I wanted the LLM to mark what it had changed, because sometimes the LLM did its job without telling me what it had changed. After thinking about it, it does increase the possibility of tool confusion and I will take your suggestion and find a better way.

PR #21, opened 2025-04-03
added linux support excluding claude

tarterp:
I've been running on linux without problem. I have verified all tools return success on IDA version `9.0sp1` and `9.1`. I am not sure if you have any other concerns beyond Claude doesn't support linux.


mrexodia:
Looks good to me, thank you!

PR #36, opened 2025-04-05
refactor MCP and servers to support sse and idalib

williballenthin:
Opening this PR to prompt discussion and design, not to merge as-is.

In these changes I've:
  - split the tools from the IDA Pro plugin script, and
  - added a script to run an MCP sse server using idalib

I have not updated the installation steps, etc. so again, this shoudn't be merged as-is.

The idea for how this works is:
  - if running in IDA Pro, put `mcp_ida_plugin.py` in $IDAUSR/plugins and then setup `remote_ida_mcp_server.py` as an stdio-based MCP server.
  - if running headless, run `python idalib_mcp_server.py /path/to/exe` as a sse-based MCP server directly.

Both modes rely on `ida_tools.py` as the core integration with the IDA API.

--- 

At the moment, I haven't tested/run all the modes, mostly just refactored a little blind and supported by linters. Again, this is to spark discussion, not a ready implementation.


williballenthin:
Things that should be done:
  - [ ] maybe move `remote_ida_mcp_server.py` to `scripts/` since its not really part of the library. counterpoint: keep it there so it can be invoked as `python -m ida_pro_mcp.remote_ida_mcp_server` after installation from PyPI, which is convenient.
  - [ ] ditto for `idalib_mcp_server.py`
  - [ ] maybe remove the JSON-RPC server from `mpc_ida_plugin.py` and delete `remote_ida_mcp_server.py`, instead running an sse MCP server directly within IDA Pro (and using something like https://github.com/supercorp-ai/supergateway if the agent framework doesn't support sse). its quick to switch to sse and would enable us to delete a lot of code.

src/ida_pro_mcp/ida_tools.py
williballenthin, line 1: this file is now specifically for the integration with the IDA API to provide tools, not the mechanism to invoke the MCP server.

src/ida_pro_mcp/ida_tools.py
```diff
@@ -75,165 +72,14 @@ def jsonrpc(func: Callable) -> Callable:
     global rpc_registry
     return rpc_registry.register(func)
 
-class JSONRPCRequestHandler(http.server.BaseHTTPRequestHandler):
```
williballenthin, line 78: this JSON-RPC code moves into a separate IDA Pro plugin file, since its not used in an idalib-based setup.

src/ida_pro_mcp/idalib_mcp_server.py
```diff
@@ -0,0 +1,77 @@
...
+
+    load_file(args.input_path)
+    
+    mcp = FastMCP("github.com/mrexodia/ida-pro-mcp", log_level="INFO", host=args.host, port=args.port)
+    for name, callable in rpc_registry.methods.items():
+        mcp.add_tool(callable, name)
+
+    try:
+        logger.info("MCP Server (sse) availabile at: http://%s:%d", mcp.settings.host, mcp.settings.port)
+        mcp.run(transport="sse")
```
williballenthin, line 67: this is really easy!

src/ida_pro_mcp/remote_ida_mcp_server.py
williballenthin, line 1: this is now the stdio-based MCP server that connects via JSON-RPC to the IDA Pro instance and exposes its tools.

mrexodia:
Generally I'm fine with the splitting up of the code on the MCP server side, because users need to `pip install` that anyway, but I would like to keep the IDA plugin in a single file without any dependencies, to make installation drag&drop/symlink. In my experience users are unwilling to install plugins that require dependencies, simply because it is too painful. I also want to avoid changing the existing filenames, because there are already users and reconfiguring is going to be a pain too.

The server won't be started if you import the plugin as a regular python module, so it should be possible to do the idalib MCP server transparently. Next week I will have some more time to look at this.

williballenthin:
ok these are helpful constraints and i'm sure it's all possible while keeping the existing structure. thanks for the feedback!

williballenthin:
superseded by #39

PR #39, opened 2025-04-09
add idalib-based MCP server

### LINKED ISSUE #20: provide an idalib-based MCP server
Some people, like me :-), want to run the server directly as a cli app, rather than an IDA plugin interactively. We can use idalib for this.

williballenthin:
This PR introduces a MCP server for IDA that uses idalib to access the IDAPython API. This means that you can access IDA's analysis on a headless system, without having to open the IDA Pro GUI and launching the plugin.

This MCP server supports both stdio and sse transports. With sse, you could run the MCP server in a container or sandbox for isolation, exposing only the TCP port. Note, however, that you must restrict access to the TCP port appropriately.

Here's Cursor accessing PMA 01-01.dll and collecting the MD5 computed by IDA:
image

### transport: stdio

1. configure your agent framework to start the MCP server, like `.cursor/mcp.json`:

```json
{
    "mcpServers": {
        "ida-pro-mcp": {
            "command": "/Users/user/code/ida-pro-mcp/.venv/bin/python",
            "args": ["-m", "ida_pro_mcp.idalib_mcp_server", "--transport=stdio", "/Users/user/code/ida-pro-mcp/pma.dll_"]
        }
    }
}
```

### transport: sse

1. start the server: `python -m ida_pro_mcp.idalib_mcp_server --transport=sse /path/to/file/to/analyze.exe`
2. configure your agent framework, like `.cursor/mcp.json`:

```json
{
    "mcpServers": {
        "ida-pro-mcp": {
            "url": "http://localhost:8000/sse"
        }
    }
}
```

## transport tradeoffs

As you can see from above, you have to specify the input file path in the MCP/stdio server configuration, which can be tedious to update. The MCP/sse server can be restarted independently of the agent software and point to different input files (which might be confusing in its own right).

So, perhaps its worth adding a tool to let the agent open the input file (restricted to a given directory, of course), so you could say "open foo.exe and determine its MD5" or something.

closes #20

### DIFF
--- a/src/ida_pro_mcp/idalib_mcp_server.py
+++ b/src/ida_pro_mcp/idalib_mcp_server.py
@@ -1,8 +1,10 @@
 import sys
+import inspect
 import logging
 import argparse
 import importlib
 from pathlib import Path
+import typing_inspection.introspection as intro
 
 from rich.console import Console
 from rich.logging import RichHandler
@@ -16,6 +18,91 @@ import ida_auto
 import ida_hexrays
 
 logger = logging.getLogger(__name__)
+
+
+def fixup_tool_argument_descriptions(mcp: FastMCP):
+    # In our tool definitions within `mcp-plugin.py`, we use `typing.Annotated` on function parameters
+    # to attach documentation. For example:
+    #
+    #     def get_function_by_name(
+    #         name: Annotated[str, "Name of the function to get"]
+    #     ) -> Function:
+    #         """Get a function by its name"""
+    #         ...
+    #
+    # However, the interpretation of Annotated is left up to static analyzers and other tools.
+    # FastMCP doesn't have any special handling for these comments, so we splice them into the
+    # tool metadata ourselves here.
+    #
+    # Example, before:
+    #
+    #     tool.parameter={
+    #       properties: {
+    #         name: {
+    #           title: "Name",
+    #           type: "string"
+    #         }
+    #       },
+    #       required: ["name"],
+    #       title: "get_function_by_nameArguments",
+    #       type: "object"
+    #     }
+    #
+    # Example, after:
+    #
+    #     tool.parameter={
+    #       properties: {
+    #         name: {
+    #           title: "Name",
+    #           type: "string"
+    #           description: "Name of the function to get"
+    #         }
+    #       },
+    #       required: ["name"],
+    #       title: "get_function_by_nameArguments",
+    #       type: "object"
+    #     }
+    #
+    # References:
+    #   - https://docs.python.org/3/library/typing.html#typing.Annotated
+    #   - https://fastapi.tiangolo.com/python-types/#type-hints-with-metadata-annotations
+
+    # unfortunately, FastMCP.list_tools() is async, so we break with best practices and reach into `._tool_manager`
+    # rather than spinning up an asyncio runtime just to fetch the (non-async) list of tools.
+    for tool in mcp._tool_manager.list_tools():
+        sig = inspect.signature(tool.fn)
+        for name, parameter in sig.parameters.items():
+            # this instance is a raw `typing._AnnotatedAlias` that we can't do anything with directly.
+            # it renders like:
+            #
+            #      typing.Annotated[str, 'Name of the function to get']
+            if not parameter.annotation:
+                continue
+
+            # this instance will look something like:
+            #
+            #     InspectedAnnotation(type=<class 'str'>, qualifiers=set(), metadata=['Name of the function to get'])
+            #
+            annotation = intro.inspect_annotation(
+                                                  parameter.annotation,
+                                                  annotation_source=intro.AnnotationSource.ANY
+                                              )
+
+            # for our use case, where we attach a single string annotation that is meant as documentation,
+            # we extract that string and assign it to "description" in the tool metadata.
+
+            if annotation.type is not str:
+                continue
+
+            if len(annotation.metadata) != 1:
+                continue
+
+            description = annotation.metadata[0]
+            if not isinstance(description, str):
+                continue
+
+            logger.debug("adding parameter documentation %s(%s='%s')", tool.name, name, description)
+            tool.parameters["properties"][name]["description"] = description
 
 
 def main() -> int:
@@ -62,7 +149,7 @@ def main() -> int:
 
     if not ida_hexrays.init_hexrays_plugin():
         raise RuntimeError("failed to initialize Hex-Rays decompiler")
-    
+
     # for MCP/stdio, STDERR should not imply an error with MCP, but some clients are buggy today.
     # see: https://github.com/modelcontextprotocol/modelcontextprotocol/issues/177
     mcp = FastMCP("github.com/mrexodia/ida-pro-mcp", log_level=log_level_s, host=args.host, port=args.port)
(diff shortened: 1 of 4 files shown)


src/ida_pro_mcp/idalib_mcp_server.py
```diff
@@ -0,0 +1,92 @@
...
+
+    logger.debug("idalib: waiting for analysis...")
+    ida_auto.auto_wait()
+
+    if not ida_hexrays.init_hexrays_plugin():
+        raise RuntimeError("failed to initialize Hex-Rays decompiler")
+    
+    # for MCP/stdio, STDERR should not imply an error with MCP, but some clients are buggy today.
+    # see: https://github.com/modelcontextprotocol/modelcontextprotocol/issues/177
+    mcp = FastMCP("github.com/mrexodia/ida-pro-mcp", log_level=log_level_s, host=args.host, port=args.port)
```
mrexodia, line 67: Not entirely sure how to deal with this. The server name from `server.py` is used as the config key for Cline/Claude/etc, but this could potentially conflict with the `idalib_mcp` server which is more meant to be started manually by the user. Perhaps we should use a separate name `idalib_mcp` to distinguish them?

To be fair I do not know if this name shows up anywhere (eg when querying the available functions) so not sure it actually matters...

src/ida_pro_mcp/idalib_mcp_server.py
```diff
@@ -0,0 +1,92 @@
...
+
+    plugin = importlib.import_module("ida_pro_mcp.mcp-plugin")
+    logger.debug("adding tools...")
+    for name, callable in plugin.rpc_registry.methods.items():
+        logger.debug("adding tool: %s: %s", name, callable)
+        mcp.add_tool(callable, name)
```
mrexodia, line 73: Does this correctly understand the `Annotate[str, "Description of the parameter"]` for parameter annotations? I saw they really like this pydantic thing and I had to convert them in the `server.py`.

src/ida_pro_mcp/idalib_mcp_server.py
```diff
@@ -0,0 +1,92 @@
...
+import ida_auto
+import ida_hexrays
+
+logger = logging.getLogger(__name__)
+
+
+def main() -> int:
+    parser = argparse.ArgumentParser(description="MCP server for IDA Pro via idalib")
+    parser.add_argument("--verbose", action="store_true", help="Show debug messages")
+    parser.add_argument("--transport", choices=["stdio", "sse"], default="stdio", help="MCP protocol to use")
```
mrexodia, line 24: What do you think about unifying this into a single parameter? `--transport stdio` vs `--transport http://127.0.0.1:8000` Probably `--transport` is not the right name in that case though.

src/ida_pro_mcp/__init__.py
mrexodia, line 1: Do we need this?

src/ida_pro_mcp/idalib_mcp_server.py
williballenthin: thats reasonable.

an alternatively might be to graduate the transport and details into a required positional argument, perhaps like:

    python -m ida_pro_mcp.idalib_mcp_server /path/to/file.exe stdio
    python -m ida_pro_mcp.idalib_mcp_server /path/to/file.exe http://localhost:8000

src/ida_pro_mcp/__init__.py
williballenthin: i believe this is idiomatic to create a "regular package" in Python. while it may work without, i think this is a side-effect of namespace packages (https://peps.python.org/pep-0420/) rather than an encouraged construct.

src/ida_pro_mcp/idalib_mcp_server.py
williballenthin: looks like no:
image

we'd want to see the string `"Name of the function to get"` in that object, but its not there, only the name (`"name"`).

I can splice this in. However, how would you feel about putting the parameter documentation directly in the docstrings, rather than using `Annotated` (which is fairly new/not yet widely supported, as you see here). For example:


```py
def get_function_by_name(name: str) -> Function:
    """Get a function by its name

    args:
      name: Name of the function to get
    """
```

today, for comparison:

```py
def get_function_by_name(
    name: Annotated[str, "Name of the function to get"]
) -> Function:
    """Get a function by its name"""
```

I believe that an LLM will be able to interpret the docstring-provided argument details as well as the `Annotation`-based ones, since most legacy Python code today uses docstrings.

src/ida_pro_mcp/idalib_mcp_server.py
mrexodia: Yeah it looks like this name is completely unused by the protocol, so let's keep it as-is.

src/ida_pro_mcp/__init__.py
mrexodia: Seems like Python found yet another way to needlessly complicate everything with impossible to read standards that nobody should have to care about 😅

src/ida_pro_mcp/idalib_mcp_server.py
mrexodia: I used `Annotated[str, "xxx"]` because it is standard Python and the most concise way to express what we want (annotate a function parameter). Docstrings is a hard no for me, they were always a stop-gap until something better was invented (which happened). The pydantic thing is just unnecessary in my opinion and `Annotated[T, str]` should work with fastmcp, I will see if I can find time to upstream support for that.

In the meantime it will require a bit of `inspect` to generate a new function with the right annotations and/or directly register the functions with the mcp via their internal mechanism.

For testing I usually use this command (just updated the readme):

```sh
uv run mcp dev src/ida_pro_mcp/server.py
```

This will give you a simple web ui that shows all the tools/resources and you can manually test. It also supports SSE but you have to change the run command in the web ui before hitting 'connect'.

src/ida_pro_mcp/idalib_mcp_server.py
mrexodia: I thought about it a bit more and perhaps it will be cleaner to let the presence of `--address` or `--port` imply SSE and without them we pick stdio?

src/ida_pro_mcp/idalib_mcp_server.py
williballenthin: I've added code to handle our interpretation of `Annotated` and register that metadata as documentation with the tool. Please see https://github.com/mrexodia/ida-pro-mcp/pull/39/commits/63c3a1935bc2641410e317d1fa32030fd8e3797c

Here's the debug log showing the metadata being parsed and added:
image

And an example of the updated tool metadata:
image

And Claude showing that it can access this information:
image

src/ida_pro_mcp/idalib_mcp_server.py
williballenthin: I find that to be a little implicit, which can be fixed either with more help output, or perhaps renaming the arguments a little more.

How about an optional argument `--transport` with default value of `--transport=stdio`, but can be overridden to be `--transport=sse://localhost:8000`. This way, the default does what most people expect, but there's a way to concisely pick SSE with configuration.

In any case, I don't feel too strongly about any of this, and am happy to go along with whatever you prefer.

mrexodia: Yeah doing `--transport=sse://localhost:8000` is good for me!

src/ida_pro_mcp/idalib_mcp_server.py
mrexodia: Awesome!

mrexodia:
Thanks a lot for your contribution and productive discussions!

Issue #40, opened 2025-04-09
Is Mac OS not supported?


waddwaw:
pip install --upgrade git+https://github.com/mrexodia/ida-pro-mcp
Collecting git+https://github.com/mrexodia/ida-pro-mcp
  Cloning https://github.com/mrexodia/ida-pro-mcp to /private/var/folders/wf/j7p8w9tx117g7wy01t7tmyv80000gn/T/pip-req-build-kg58no84
  Running command git clone --filter=blob:none --quiet https://github.com/mrexodia/ida-pro-mcp /private/var/folders/wf/j7p8w9tx117g7wy01t7tmyv80000gn/T/pip-req-build-kg58no84
  Resolved https://github.com/mrexodia/ida-pro-mcp to commit 8f43326126b68f4d50d055871688f75aba4de687
  Installing build dependencies ... done
  Getting requirements to build wheel ... error
  error: subprocess-exited-with-error
  
  × Getting requirements to build wheel did not run successfully.
  │ exit code: 1
  ╰─> [92 lines of output]
      configuration error: `project.license` must be valid exactly by one definition (2 matches found):
      
          - keys:
              'file': {type: string}
            required: ['file']
          - keys:
              'text': {type: string}
            required: ['text']
      
      DESCRIPTION:
          `Project license <https://peps.python.org/pep-0621/#license>`_.
      
      GIVEN VALUE:
          "MIT"
      
      OFFENDING RULE: 'oneOf'
      
      DEFINITION:
          {
              "oneOf": [
                  {
                      "properties": {
                          "file": {
                              "type": "string",
                              "$$description": [
                                  "Relative path to the file (UTF-8) which contains the license for the",
                                  "project."
                              ]
                          }
                      },
                      "required": [
                          "file"
                      ]
                  },
                  {
                      "properties": {
                          "text": {
                              "type": "string",
                              "$$description": [
                                  "The license of the project whose meaning is that of the",
                                  "`License field from the core metadata",
                                  "<https://packaging.python.org/specifications/core-metadata/#license>`_."
                              ]
                          }
                      },
                      "required": [
                          "text"
                      ]
                  }
              ]
          }
      Traceback (most recent call last):
        File "/opt/homebrew/lib/python3.12/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 389, in <module>
          main()
[...]

PR #42, opened 2025-04-09
pyproject: add MIT license classifier

williballenthin:
ref #40 and [27c6ccc](https://github.com/mrexodia/ida-pro-mcp/commit/27c6ccc233834c8c1d448d7999bb353b723698ba)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,6 +13,7 @@ classifiers = [
     "Programming Language :: Python :: 3.11",
     "Operating System :: MacOS",
     "Operating System :: Microsoft :: Windows",
+    "License :: OSI Approved :: MIT License",
 ]
 dependencies = [
     "mcp>=1.6.0",


mrexodia:
Yeah I removed this before because I was getting warnings that we should be using the `license` field in the latest pyproject.toml standard. For me it's fine to just not have the correct license in pip, as long as it works. I would wait for feedback from #40 before trying this.

Rant: I honestly really hate this new `pyproject.toml`, nothing ever works and they always decide to completely break everything between pip/python versions. I actually spend almost 10 hours trying to get the package to install correctly with python files in the root, but it's actually impossibly broken. In the end I gave up and created a useless `src/ida_pro_mcp` folder for no actual good reason...

mrexodia:
From the report it currently works, so I will close this to avoid any further potential breakages.

image

The license is in the wheel file, so I guess uv/pyproject/whatever picked it up from the filename(?)

PR #46, opened 2025-04-10
Add debugger mpc function

ltlly:
Add debugger-related mpc function, including functionality to retrieve all registers, call stack, byte values, strings, and more.

### DIFF
--- a/src/ida_pro_mcp/mcp-plugin.py
+++ b/src/ida_pro_mcp/mcp-plugin.py
@@ -1,3 +1,4 @@
+import os
 import sys
 
 if sys.version_info < (3, 11):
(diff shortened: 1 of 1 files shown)


mrexodia:
Thanks for your contribution! As-is I do not want to merge it for a few reasons:

1. All the code was reformatted, which I do not want. PEP8 is wrong 😅 
2. Adding debugger support opens the MCP up to RCE exploits with prompt injection. If we want to support this it will need to be explicitly enabled when starting the MCP server, so that requires additional design work.
3. Many functions unrelated to the debugger are also added, please split them into a separate PR.

mrexodia:
Why did you close it? I'm happy to incorporate the changes myself.

mrexodia:
Thanks!

PR #47, opened 2025-04-13
Add functionality to list globals

neoz:


mrexodia:
Do you have an example chat where these functions were actually useful?

neoz:
> Do you have an example chat where these functions were actually useful?

In case the import search info in names list (analysis by IDA). Example:
- Analyze vtable with name xxx::LicenseInfo::`vftable',examine the function pointers and decompile each one.
- Inspect the decompilation and add comments with your findings.

mrexodia:
Thanks!

privacyguy123:
Seems memory functionality was added and removed before I even got a chance to check it out - what was the thought process behind this?

mrexodia:
Some unresolved design questions, see #54

PR #51, opened 2025-04-18
Add a function to execute idapython

b3nguang:


mrexodia:
https://x.com/junr0n/status/1905978316043219198

mrexodia:
I am currently in the process of adding an `@unsafe` specifier that will disable functions like this per default.

Issue #54, opened 2025-04-18
Add memory functionality


mrexodia:
The difficulty is designing an API that will be useful. Just returning bytes seems prone to hallucinations, but there are definitely valid use cases (analyzing vtables #47). Here is some code from #46:

```
@jsonrpc
@idaread
def data_read_byte(
    address: Annotated[str, "Address to get 1 byte value from"],
) -> int:
    """Get 1 byte value at the specified address"""
    ea = parse_address(address)
    return ida_bytes.get_wide_byte(ea)

@jsonrpc
@idaread
def data_read_word(
    address: Annotated[str, "Address to get 2 bytes value from"],
) -> int:
    """Get 2 bytes value at the specified address"""
    ea = parse_address(address)
    return ida_bytes.get_wide_word(ea)

@jsonrpc
@idaread
def data_read_dword(
    address: Annotated[str, "Address to get 4 bytes value from"],
) -> int:
    """Get 4 bytes value at the specified address"""
    ea = parse_address(address)
    return ida_bytes.get_wide_dword(ea)

@jsonrpc
@idaread
def data_read_qword(address: Annotated[str, "Address to get 8 bytes value from"]) -> int:
    """Get 8 bytes value at the specified address"""
    ea = parse_address(address)
    return ida_bytes.get_qword(ea)

@jsonrpc
@idaread
def data_read_string(address: Annotated[str, "Address to get string from"]):
    """Get string at the specified address"""
    try:
        return idaapi.get_strlit_contents(parse_address(address),-1,0).decode("utf-8")
    except Exception as e:
        return "Error:" + str(e)

@jsonrpc
@idaread
def get_memory_bytes(
    address: Annotated[str, "Address to get memory bytes from"],
    size: Annotated[int, "Number of bytes to read"]
) -> list[int]:
    """Get memory bytes (hex) at the given address"""
    address = parse_address(address)
    data = idc.get_bytes(address, size)
    if not data:
        raise IDAError(f"Failed to read memory at {hex(address)}")
    return list(data)
```

I lean towards `data_read_uint8`, `data_read_uint16`, `data_read_uint32`, `data_read_uint64` that return `hex(value)` and discuss `convert_number` in their descriptions. Reading a string is also fine, but `get_memory_bytes` seems rather pointless (unless the bytes need to be extracted into a script or something).

withzombies:
I missed this issue before working on my PR #82. I went with slightly different route, which is to implement it as two functions: 
[...]

PR #57, opened 2025-04-22
Add Patch Address bytecode

ShallowFeather:
I notice that it can't patch the assembly, so I add it.
It can be used to deobfuscate some simple obfuscate, like the sample in 

https://github.com/user-attachments/assets/560c7528-69f5-4e7a-8c45-5d9fab36e452

### DIFF
--- a/src/ida_pro_mcp/mcp-plugin.py
+++ b/src/ida_pro_mcp/mcp-plugin.py
@@ -885,6 +885,23 @@ def set_global_variable_type(
 
 @jsonrpc
 @idawrite
+def patch_address_bytecode(
+    address: Annotated[str, "Address of the instruction to patch"],
+    bytecode: Annotated[str, "Array of bytecode to patch the instruction with"],
+) -> str:
+    """Patch Address Assemble"""
+    ea = int(address, 16)
+    byte_list = bytecode.strip().split()
+    bytes_to_patch = bytes(int(b, 16) for b in byte_list)
+    try:
+        ida_bytes.patch_bytes(ea, bytes_to_patch)
+    except:
+        raise IDAError(f"Failed to patch bytes at address {hex(ea)}")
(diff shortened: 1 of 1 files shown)


mrexodia (changes requested):
Thanks for your contribution! Some changes will be necessary before merging.

mrexodia:
From watching the video it looks like this relies on the LLM to automagically generate instruction bytes. I am a bit hesitant to merge functionality like this, because it is prone to hallucinations. You would have to use another MCP to assemble using keystone and get the bytes that way.

ShallowFeather:
I'm not entirely sure if this project would want to introduce external libraries like Keystone for implementation, which is why I directly used bytecode as input. Initially, I was looking into whether IDApython had an interface to achieve this functionality, but I only found methods for patching by inputting bytecode.

mrexodia:
No, the IDA plugin should not have any dependencies. I think it should be part of #54 as `write_bytes`, but the design is still a bit unclear to me…

Another option is to add keystone as a dependency to the MCP server itself, but I’m not the biggest fan of that either. Will think about it and get back to you!

In the meantime I suggest you code up a simple keystone-mcp, I can link it in the readme.

williballenthin:
could you use `idautils.Assemble()`? https://python.docs.hex-rays.com/namespaceidautils.html#a9bd14bb8029f2e7e7ce8715fa3d60638

or

`ida_idp.AssembleLine` 
https://python.docs.hex-rays.com/namespaceida__idp.html#afcd433111d5f7169b516752965491ec9

ShallowFeather:
Thanks @williballenthin, I edited it to `patch_address_assemble`
But I notice that the patch will not be only one line maybe have to add function for multiline patch

mrexodia:
Thanks!

PR #63, opened 2025-04-30
feat: Add support for latest Python versions while maintaining backward compatibility

seifreed:
- Expand Python version compatibility to include 3.8-3.12 (previously only >=3.11)  
- Add Python 3.8, 3.9, 3.10, and 3.12 classifiers in pyproject.toml
- Modify version check in mcp-plugin.py to accept Python 3.8+
- Update .python-version to 3.12 to test with latest Python
- Add typing_extensions dependency for Annotated type compatibility in older versions
- Implement conditional imports for Annotated type across all files
- Fix TypedDict and Generic compatibility issue in Python 3.10
- Update README to accurately reflect the wider Python version compatibility
- Fix AST processing in server.py for better cross-version support

### DIFF
--- a/src/ida_pro_mcp/mcp-plugin.py
+++ b/src/ida_pro_mcp/mcp-plugin.py
@@ -599,11 +599,25 @@ def convert_number(
 
 T = TypeVar("T")
 
-class Page(TypedDict, Generic[T]):
-    data: list[T]
+# In Python 3.10, TypedDict cannot be used with Generic directly
+# Using a different approach for compatibility
+class PageDict(TypedDict):
+    data: list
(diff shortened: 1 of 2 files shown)


mrexodia:
Why do you need to support older Python versions?

mrexodia:
According to https://pypi.org/project/mcp/ the `mcp` dependency has a minimum supported Python version of 3.10, does it actually work with 3.8?

mrexodia:
🤔

seifreed:
Sorry! Wrong commit! I'll do another soon! Thanks for the job you do :)

PR #70, opened 2025-05-14
add local type listing and searching function

hoooonnam:
I added a function to get list of local types in file loaded in IDA.


mrexodia:
Thanks for your contribution!

Trass3r:
Does this + `declare_c_type` enable the model to refine data structures, at least in theory?

mrexodia:
Yeah it should! The `declare_c_type` overwrites an existing type if you make one with the same name if I remember correctly from my experiments.

PR #80, opened 2025-06-02
Allow set_comment to work with no decompiler available

withzombies:
Currently, if you do not have a decompiler for the platform, the `set_comment` function fails.

This PR adds a check to see if the hex rays plugin has been initialized and only sets the function comment if not.

### DIFF
--- a/src/ida_pro_mcp/mcp-plugin.py
+++ b/src/ida_pro_mcp/mcp-plugin.py
@@ -275,6 +275,9 @@ class IDAError(Exception):
 class IDASyncError(Exception):
     pass
 
+class DecompilerLicenseError(IDAError):
+    pass
+
 # Important note: Always make sure the return value from your function f is a
 # copy of the data you have gotten from IDA, and not the original data.
 #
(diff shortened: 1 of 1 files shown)


src/ida_pro_mcp/mcp-plugin.py
```diff
@@ -891,9 +891,15 @@ def set_comment(
     if not idaapi.set_cmt(address, comment, False):
         raise IDAError(f"Failed to set disassembly comment at {hex(address)}")
 
+    if not ida_hexrays.init_hexrays_plugin():
+        return
+
     # Reference: https://cyber.wtf/2019/03/22/using-ida-python-to-analyze-trickbot/
     # Check if the address corresponds to a line
-    cfunc = decompile_checked(address)
+    try:
```
mrexodia, line 899: This will hide a real decompilation error. Did you observe `ida_hexrays.init_hexrays_plugin()` succeeding but the decompilation failing if you do not have a decompiler available?

mrexodia (changes requested):
Thanks a lot! Added a small comment/question.

src/ida_pro_mcp/mcp-plugin.py
withzombies: If you have a HexRays decompiler but not the right decompiler for the executable , the init function still returns true. 

For example, I hit this when reversing an x86 binary but I only have the x64 and ARM64 decompilers.

withzombies: I figured out a way to specifically check for license problems. I introduced a new error type called `DecompilerLicenseError`, which I check in this try now. The other decompilation errors should bubble up.

mrexodia:
Thanks a lot for your contributions!

Issue #52, opened 2025-04-18
More verbose error reporting


privacyguy123:
Maybe we can let the AI Agent know it tried to rename something and it was a duplicate?

```
6ACD8: can't rename byte as 'FT_MulFix' because the name is already used in the program.
```

privacyguy123:
Adding more examples - when the AI Agent uses "get_xrefs_to" and it's empty it will continually use the comand in a loop because the error message is not stating the function has no xrefs.

kruxmeier:
https://github.com/mrexodia/ida-pro-mcp/blob/2f0f925adbca7104789a0930ce0d76d9b6897aab/src/ida_pro_mcp/mcp-plugin.py#L738-L741
Does not give the actual error, `f"Decompilation failed at {hex(address)}: {error}"` is better.
Similarly you should add `: {e}` here: https://github.com/mrexodia/ida-pro-mcp/blob/2f0f925adbca7104789a0930ce0d76d9b6897aab/src/ida_pro_mcp/mcp-plugin.py#L1015
Setting the prototype still sometimes fails without proper explanation, e.g. I guess if types mentioned in the prototype are not defined yet.

PR #82, opened 2025-06-03
Add memory read functions

withzombies:
Adds two functions to read memory values. Without the decompiler, the LLM was having trouble figuring out values, since they're not always inlined in the disassembly.

Screenshot 2025-06-03 at 5 01 26 PM

### DIFF
--- a/src/ida_pro_mcp/mcp-plugin.py
+++ b/src/ida_pro_mcp/mcp-plugin.py
@@ -981,6 +981,36 @@ def set_global_variable_type(
         raise IDAError(f"Failed to apply type")
 
 @jsonrpc
+@idaread
+def get_global_variable_value(variable_name: Annotated[str, "Name of the global variable"]) -> str:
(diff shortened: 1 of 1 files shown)


mrexodia:
I would like to defer this until I have time to properly look at #54 and decide on an API pattern that works for every case.

mrexodia:
I didn't have much time to play with the MCP server recently, but it looks like you have been actively working with it @withzombies. Do you think this is ready for merging, or should we adjust the function descriptions as you mentioned in #52 ?

withzombies:
I've also been out for the past two weeks+, so I haven't poked at it much. I think I'd like to update the docstrings, my feeling is adding extra context to the docstrings should only help.

I'll set some time aside soon to update the docstrings and maybe introduce a `get_global_value_at(address)` function.

withzombies:
I think this is ready to merge

mrexodia:
Awesome, thanks!

PR #88, opened 2025-06-16
Add get variable(ex. globals variable) values by its name or address

hoooonnam:
this function Retrieve the value of a variable within any range you want by its name or address. Can be helpful when solving CTF challenges , vulnerability research and etc.
get_variable_by_name
get_variable_by_address


mrexodia:
Seems like a duplicate of #82 (discussion in #54).

PR #123, opened 2025-08-23
Add comprehensive structure analysis functionality

VadimBoev:
in IDA PRO has a tab called "Local Types", where you can find structures (either created or from binary file analysis).
For example, there is such a structure:

image

Let's ask any AI to find us this structure (the whole structure didn't fit in the screenshot):
image

### DIFF
--- a/src/ida_pro_mcp/mcp-plugin.py
+++ b/src/ida_pro_mcp/mcp-plugin.py
@@ -1526,6 +1526,201 @@ def get_defined_structures() -> list[StructureDefinition]:
     return rv
 
 @jsonrpc
+@idaread
+def analyze_struct_detailed(name: Annotated[str, "Name of the structure to analyze"]) -> dict:
+    """Detailed analysis of a structure with all fields"""
+    # Get tinfo object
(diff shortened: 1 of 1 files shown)


VadimBoev:
I've never touched IDA PRO from the Python scripting side, but this documentation helped me in this process (maybe it will be useful to someone) - [porting guide from 8.x to 9.x](https://docs.hex-rays.com/developer-guide/idapython/idapython-porting-guide-ida-9)

mrexodia:
Not against adding it per se, but what is the actual use case for this? How does it help the LLM reverse engineer in a practical sense?

VadimBoev:
> Not against adding it per se, but what is the actual use case for this? How does it help the LLM reverse engineer in a practical sense?

In my practice, I have already encountered when I needed to write a modification for a game, but I made mistakes in the structures inside the C/C++ code. With this implementation, checking structures is much easier and more reliable (saves time). In addition, it allows you to better see the 'pseudo-code' if it was poorly decompiled (which also saves time).

mrexodia:
Thank you!

Issue #132, opened 2025-09-16
set_comment failed because of  Decompilation failed


heheda123123:
There are many functions that IDA cannot decompile, so setting comments will fail.
Setting comments should have nothing to do with decompilation, right? Or when decompilation is not possible, it will fall back to assembly for analysis?    

<img width="1906" height="623" alt="Image" src="https://github.com/user-attachments/assets/cc890bfb-5093-4447-bc54-766e8d550e31" />

This PR mentioned the relevant situation, but the problem here is that ida has a decompilation function, but it cannot decompile the current function
https://github.com/mrexodia/ida-pro-mcp/pull/80

heheda123123:
My understanding is wrong, this should be because ida has different types of comments 
The several types of comments in ida are not related. I think it is possible to add a parameter to set_comment to determine whether to add comments in the disassembly window or the decompilation window.
in the disassembly window can use `:` add comment
int the decompilation window can use `/` add comment

heheda123123:
Or replace the current `set_comment` with two other functions `set_comment_in_disassembly` `set_comment_in_decompilation` This may be more intuitive

PR #133, opened 2025-09-16
modify set_comment

heheda123123:
solve the problem in this issue
already test the two new function 
https://github.com/mrexodia/ida-pro-mcp/issues/132


mrexodia:
I do not agree with this approach, we need fewer functions not more of them.

heheda123123:
> I do not agree with this approach, we need fewer functions not more of them.

Do you think adding parameters to set_comment is a viable approach?

PR #134, opened 2025-09-16
modify set_comment

heheda123123:
https://github.com/mrexodia/ida-pro-mcp/issues/132
https://github.com/mrexodia/ida-pro-mcp/pull/133

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ Available functionality:
 - `get_xrefs_to(address)`: Get all cross references to the given address.
 - `get_xrefs_to_field(struct_name, field_name)`: Get all cross references to a named struct field (member).
 - `get_entry_points()`: Get all entry points in the database.
-- `set_comment(address, comment)`: Set a comment for a given address in the function disassembly and pseudocode.
+- `set_comment(address, comment, in_decompile_window)`: Set a comment for a given address in the function disassembly or pseudocode.
 - `rename_local_variable(function_address, old_name, new_name)`: Rename a local variable in a function.
 - `rename_global_variable(old_name, new_name)`: Rename a global variable.
 - `set_global_variable_type(variable_name, new_type)`: Set a global variable's type.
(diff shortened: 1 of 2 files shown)


mrexodia:
This has exactly the same issue as the PR I just closed though? The intention of `set_comment` is that is _transparently_ sets a comment. It should put the comment on the disassembly and the decompilation, whatever is possible.

heheda123123:
> This has exactly the same issue as the PR I just closed though? The intention of `set_comment` is that is _transparently_ sets a comment. It should put the comment on the disassembly and the decompilation, whatever is possible.

However, in IDA, the comments in the disassembly window and the decompilation window are not interoperable. Moreover, the number of lines of code in the decompiled code does not completely correspond to the decompiled code. For example, 3 lines of assembly code correspond to one line of pseudo code. I add comments to each line of assembly code. How can I add them to the pseudo code?

mrexodia:
It's not very important in my opinion. The goal of the MCP server is to do the right thing and if you set a comment it should do a best-effort to set it in both the disassembly and the decompilation. The main work is done with the decompilation, so setting a comment there is the most common action. Every line has a corresponding address in the API so setting a comment there should set it in both the decompilation at that line as well as the disassembly at that address. If you set a comment in the disassembly at an address that does not have a line just don't set it in the decompilation?

heheda123123:
> It's not very important in my opinion. The goal of the MCP server is to do the right thing and if you set a comment it should do a best-effort to set it in both the disassembly and the decompilation. The main work is done with the decompilation, so setting a comment there is the most common action. Every line has a corresponding address in the API so setting a comment there should set it in both the decompilation at that line as well as the disassembly at that address. If you set a comment in the disassembly at an address that does not have a line just don't set it in the decompilation?

But there are two situations, one is the obfuscation situation, the other is the pure assembly code

mrexodia:
Okay, but if the decompilation fails the idea is to just use the disassembly instead?

mrexodia:
I fixed it by returning early if decompilation fails for any reason.

PR #138, opened 2025-09-28
Added 2 new tools

MateiMartin:


mrexodia:
Thanks, I integrated your changes to `get_disassembly`. I do not see any reason to add capa to the project, this should be a separate MCP.

Issue #89, opened 2025-06-17
Support for multiple IDA instances


zhangdafeihhh:
When some libs are to be analysed together, two or more ida/plugins are needed to listen on multiples ports

luodaoyi:
me too

Haicaji:
I'm considering whether it's possible to control multiple instances by running services on different ports. See PR #143

mrexodia:
There has been a major overhaul. You can now start up to 10 instances as implemented in #143 by @Haicaji, but the database switching has not been implemented yet. It would require modifying [`server.py`](https://github.com/mrexodia/ida-pro-mcp/blob/main/src/ida_pro_mcp/server.py#L31) to allow combining local `@tool` decorators with the tools available from IDA. Unfortunately clients do not support refreshing the tool list, so at least one client needs to be running for this to work.

Some other ideas:
- Make the `idalib` MCP a first-class citizen (headless), with this you could call `open_database("my.exe")` as well as enumerate running IDA instances and provide `list_instances` and `switch_instance`
- Write a tool to automatically convert an MCP server to a command line utility (+ a Claude skill)
  - https://blog.fsck.com/2025/10/19/mcps-are-not-like-other-apis/
  - https://mariozechner.at/posts/2025-11-02-what-if-you-dont-need-mcp/?t=0

NeKroFR:
I think we can close this issue @mrexodia

PR #143, opened 2025-10-07
Support for multiple instances

Haicaji:
This PR originates from issue #89 .

I've implemented a feature where the IDA MCP plugin (mcp-plugin.py), upon startup, automatically finds and uses an available port within a specific range (13337-13437).

Subsequently, on the server side, I have provided the functions list_ida_instances() and switch_ida_instance(port) to display the current instances and to switch control between them.

This should be a relatively straightforward way to control multiple IDA instances. However, this approach requires the large model to have strong contextual capabilities, and perhaps there is a better way.

### DIFF
--- a/src/ida_pro_mcp/mcp-plugin.py
+++ b/src/ida_pro_mcp/mcp-plugin.py
@@ -202,12 +202,28 @@ class MCPHTTPServer(http.server.HTTPServer):
 
 class Server:
     HOST = "localhost"
-    PORT = 13337
+    BASE_PORT = 13337
+    MAX_PORT_ATTEMPTS = 100  # Try ports 13337-13436
 
     def __init__(self):
         self.server = None
         self.server_thread = None
         self.running = False
+        self.port = None
+
+    def _find_available_port(self):
+        """Find an available port starting from BASE_PORT"""
+        import socket
+        for port_offset in range(self.MAX_PORT_ATTEMPTS):
+            port = self.BASE_PORT + port_offset
+            try:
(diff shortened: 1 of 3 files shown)


src/ida_pro_mcp/mcp-plugin.py
```diff
@@ -229,17 +245,83 @@ def stop(self):
...
+                current_title = None
+                
+                # Method 1: Try get_window_title (IDA 7.6+)
+                if hasattr(ida_kernwin, 'get_window_title'):
+                    current_title = ida_kernwin.get_window_title()
+                
+                # Method 2: Try via QWidget (if Qt is available)
+                if not current_title:
+                    try:
+                        from PyQt5.QtWidgets import QApplication
```
mrexodia, line 282: Newer IDA versions no longer support PyQt5, so this will need some extra work.

Haicaji: I’m not very familiar with this aspect, so I’ve decided to temporarily remove the feature that displays the port in the title. The port can still be output in the console.
Additionally, I found that PyQt5 is also used in this part.
https://github.com/mrexodia/ida-pro-mcp/blob/b62b84946e72ac870b74d1054fa4a8f753f5e43f/src/ida_pro_mcp/mcp-plugin.py#L393

mrexodia:
Changes from #170 make this unnecessary.

AudriusButkevicius:
Sorry for ignorance, how does SSE solve the issue of needing to give model access to multiple instances?

PR #149, opened 2025-10-10
chore: Enforce LF line endings via .gitattributes

bukowa:


mrexodia:
You should use `git config --global core.autocrlf false` on your system and never let git change line endings automatically. This is a very frustrating default in git that has been fucking everyone over for years, but nothing we can do...

bukowa:
> You should use `git config --global core.autocrlf false` on your system and never let git change line endings automatically. This is a very frustrating default in git that has been fucking everyone over for years, but nothing we can do...

I think with this commit it will be handled automatically by .gitattributes?

mrexodia:
I do not want to create a `.gitattributes` file reinforcing a bad default. And your change does the exact opposite and forces `core.autocrlf true`, showing that it is inherently confusing 😅

bukowa:
> I do not want to create a `.gitattributes` file reinforcing a bad default. And your change does the exact opposite and forces `core.autocrlf true`, showing that it is inherently confusing 😅

Fk because it is confusing all over the place :D
Can .gitattribute have proper defaults?

I never had such problems in other repors tbh and I always use my settings. *that would be a lie I had it once or twice but I fixed it with .gitattributes I guess

mrexodia:
The only correct value for `core.autocrlf` is `false`. If the repository wants to enable conversions they can set that in `.gitattributes`. I do not want conversions, so nothing should have to be defined. `core.autocrlf=false` is the default for git, but for unknown reasons the git for windows _installer_ sets it to something else. I cannot change what the git for windows maintainers decided, but I can protest by not allowing this silliness to propagate.

You can fix the problem (for all repositories you contribute to ever) by running the following command:

```
git config --global core.autocrlf false
```

If you don't want to change a global setting you can omit `--global` to change it just for your local copy of this repository.

bukowa:
> ```
> git config --global core.autocrlf false
> ```

Thanks for the explanation btw that is a very solid argument/
I can't tell how much time I've wasted by docker not properly reading my CLRF entrypoints

mrexodia:
Yep, Docker (devcontainers specifically) was another **major** pain point when dealing with this. I would like to do more research to see if https://github.com/git-for-windows/git/issues/2994 can be reopened and changed. The discussion there completely seems to miss how frustrating this setting is for maintainers and contributors alike.

PR #157, opened 2025-10-26
Fix headless mode startup on Python 3.11

glzjin:
Fixes Pydantic compatibility issue by using typing_extensions.TypedDict instead of typing.TypedDict on Python < 3.12, as required by Pydantic.

- Use conditional imports for TypedDict and NotRequired based on Python version
- Add typing_extensions as conditional dependency for Python < 3.12

### DIFF
--- a/src/ida_pro_mcp/mcp-plugin.py
+++ b/src/ida_pro_mcp/mcp-plugin.py
@@ -13,15 +13,20 @@ from typing import (
     Any,
     Callable,
     get_type_hints,
-    TypedDict,
     Optional,
     Annotated,
     TypeVar,
     Generic,
-    NotRequired,
     overload,
     Literal,
 )
+
+# Use typing_extensions for TypedDict and NotRequired on Python < 3.12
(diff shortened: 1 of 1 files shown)


src/ida_pro_mcp/mcp-plugin.py
```diff
@@ -13,16 +13,21 @@
...
     overload,
     Literal,
 )
 
+# Use typing_extensions for TypedDict and NotRequired on Python < 3.12
+# to ensure compatibility with Pydantic
+if sys.version_info >= (3, 12):
+    from typing import TypedDict, NotRequired
+else:
+    from typing_extensions import TypedDict, NotRequired
```
mrexodia: The `typing_extensions` package is not available inside of IDAPython per default, so this would conflict with the main use case. Such hacks should instead be in `idalib_server.py`

mrexodia:
I created https://github.com/mrexodia/ida-pro-mcp/issues/171, this fix is not right.

PR #161, opened 2025-11-06
Add filter for function listing

imnuke:
There are `list_globals_filter`, `list_strings_filter`, but no `list_functions_filter`, I added it.
In some big binaries the function list can be long, checking all of them will consume many tokens. Sometimes we only wanna focus on functions with specific patterns.


mrexodia:
Thanks!

PR #162, opened 2025-11-06
Fix IDA database closing on idalib server shutdown

imnuke:
The IDA database was not being properly closed when the MCP server received SIGINT (Ctrl+C) or SIGTERM signals, leaving the `.id0`, `id1`, `.nam`, `.til` files and there might be potential data loss. 

The original code attempted to handle this with a `try-except KeyboardInterrupt` block, but this never executed.

This occurred because uvicorn re-raises signals after graceful shutdown, immediately terminating the process before cleanup code could execute.

Solution: Install signal handlers for SIGINT/SIGTERM that close the database before exiting. Also added atexit handler as backup for normal exits.


mrexodia (changes requested):
.

mrexodia:
Thanks!

Issue #163, opened 2025-11-07
the `list_functions` in the main branch is causing the entire server to freeze.


Ps7ch3:
When I start MCP with the latest commit from the main branch (commit: 10bcbd0), I noticed that calling list_functions causes the entire server to freeze. However, when I revert to version v1.4.0, everything works as expected.

This issue might be related to this change: [commit link](https://github.com/mrexodia/ida-pro-mcp/commit/d8e610632e5e4face1d9cf6475878819b5fd607c).

From the debug logs, it appears that after calling list_functions, the flow doesn’t reach list_functions_filter, and instead, the server freezes entirely.

If you need further details to reproduce the issue, feel free to contact me.

IDA Version: 9.0.240807 (Linux)
Test Tool: [Inspector Tool](https://github.com/modelcontextprotocol/inspector)

Ps7ch3:
<img width="1714" height="656" alt="Image" src="https://github.com/user-attachments/assets/2e069883-f0b1-4fa6-9c0d-cd1622cb7713" />

PR #164, opened 2025-11-07
fix: remove `idaread` from list_functions to prevent server freeze.

Ps7ch3:
See https://github.com/mrexodia/ida-pro-mcp/issues/163


mrexodia:
Good catch, thanks! I opened #165 to fix this properly.

Issue #166, opened 2025-11-08
Exception in dbg_get_registers : ValueError: Conversion failed


Spl3en:
Duplicate with https://github.com/mrexodia/ida-pro-mcp/issues/146

Version : 844b9fa7a49350d8b1db994f3ce1cde06b07c4e8 (latest)

<img width="801" height="400" alt="Image" src="https://github.com/user-attachments/assets/4b29025f-5b0c-4b66-af38-a1bd6d9bc48b" />

Here's the binary : https://files.web.dynu.net/GjAL5NwdeOBb.zip

Apparently, the blind fix (a76aaf34fa804d063b559fce27b9b54e4023ded3) didn't prevent the bug.

mrexodia:
~~Only one problem: line 1165 in your exception actually contains `if not ida_hexrays.init_hexrays_plugin():`~~ reading is hard

The `return dbg` statement cannot really throw an exception and I cannot reproduce this locally. Maybe you can try to add some try/except clauses and/log logging to see when this happens?

PR #167, opened 2025-11-08
Fix regval_t.pyval(dtype) error

Spl3en:
regval_t.pyval(dtype) throws a ValueError when trying to convert floating-point and vector registers (FPU ST0-ST7, MMX, XMM, YMM) because their internal representation (rvtype == RVT_FLOAT) doesn't match the conversion expected by pyval().

The fix catches this exception and falls back to checking the rvtype field to either use rv.ival for integer types or rv.bytes() for floating-point/custom types, displaying them as hex bytes.

### DIFF
--- a/src/ida_pro_mcp/mcp-plugin.py
+++ b/src/ida_pro_mcp/mcp-plugin.py
@@ -1986,11 +1986,23 @@ def dbg_get_registers() -> list[ThreadRegisters]:
             rv: ida_idd.regval_t
             reg_info = dbg.regs(reg_index)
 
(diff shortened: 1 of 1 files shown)


Spl3en:
See https://github.com/mrexodia/ida-pro-mcp/issues/166 for discussion

mrexodia:
This fix doesn't make much sense though, in #166 it shows that the `ValueError` exception escapes, so why would adding additional logic in the existing exception handler do anything? The exception has to be raised from a different place...

PR #172, opened 2025-11-14
Add Codex support + fix atomic config writes + Chunked http support

can1357:
**Fixes:**
- Atomic config writes (temp file + `os.replace`) to prevent corruption during concurrent
Claude Code sessions
- Resolves the issue where concurrent sessions would break `.claude.json` with partial
JSON
- HTTP server now parses chunked requests correctly, so streaming MCP transports stop
timing out during initialization

**New:**
- Codex MCP client support with TOML config (`~/.codex/config.toml`)
- All clients now use HTTP transport (`type: "http"`, `url: "http://127.0.0.1:13337/mcp"`)
instead of subprocess
- Added `tomli-w` dependency for TOML serialization

**Changes:**
- `install_mcp_servers()` now handles both JSON and TOML formats
- TOML clients use `[mcp_servers.<name>]` tables
- Removed subprocess overhead - clients connect directly to IDA plugin's HTTP endpoint

### DIFF
--- a/src/ida_pro_mcp/idalib_server.py
+++ b/src/ida_pro_mcp/idalib_server.py
@@ -1,6 +1,5 @@
 import sys
 import signal
-import atexit
 import inspect
 import logging
 import argparse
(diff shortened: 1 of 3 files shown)


mrexodia:
❤️

PR #173, opened 2025-11-14
 Add --allow-ida-free option

momo5502:
This option permits installing the plugin if IDA Free and Pro are used side by side.


mrexodia:
This was explicitly added because people kept ignoring any and all command line output and opening issues that IDA Free doesn't work. You can add a command line option `--allow-ida-free` because this seems like an extremely niche use case

mrexodia:
Thanks!

PR #176, opened 2025-11-15
Cache idautils.Strings to avoid triggering build_strlist on every call

kkHAIKE:
In large-scale projects, calling list_strings_filter triggers build_strlist every time, which is very slow.


mrexodia:
All the code was moved around, closing this.

PR #199, opened 2025-11-24
fix(disasm): start from requested address and include comments

can1357:
## Summary
- `disasm` now starts from the requested address instead of always starting from the function beginning when the address is within a function
- Switched to `generate_disasm_line` API which includes IDA comments in the output instead of manually building from mnemonic + operands

## Test plan
- [x] Call `disasm("0x441E99")` where address is mid-function - should start from that address, not function start
- [x] Comments added in IDA should appear in disasm output


mrexodia:
Bro made a nice test plan 😆

Issue #155, opened 2025-10-21
IDA Home support ?


fredericGette:
Hello,

The README indicates "IDA Free" is not supported. But what about "IDA Home" ?

By curiosity, I tried with IDA Home 9.2 and Python 3.14  
and it is partially working:

| Tool | Result |
|---|---|
| `check_connection` | Success |
| `get_metadata` | Success |
| `get_function_by_name` | Success |
| `get_function_by_address` | Success |
| `get_current_address` | Success |
| `get_current_function` | Success |
| `convert_number` | Success |
| `list_functions` | Success |
| `list_globals_filter` | Success |
| `list_globals` | Success |
| `list_imports` | Success |
| `list_strings_filter` | Success |
| `list_strings` | Success |
| `list_local_types` | Success |
| `decompile_function` | Failure (timeout) |

The `decompile_function` makes IDA Home crashes:

<img width="461" height="276" alt="Image" src="https://github.com/user-attachments/assets/2e438348-dc07-4f03-889a-7cae0fa9bdc4" />

And the logs are the following:
```
-----------------------------------------------------------------------------------------
Python 3.14.0 (tags/v3.14.0:ebf955d, Oct  7 2025, 10:15:03) [MSC v.1944 64 bit (AMD64)] 
IDAPython 64-bit v9.2.0 (c) The IDAPython Team <idapython@googlegroups.com>
-----------------------------------------------------------------------------------------
[MCP] Server started at http://localhost:13337
419008: restored microcode from idb
419008: restored pseudocode from idb
419066: variable 'v5' is possibly undefined
40E3B8: using guessed type _EVENT_DESCRIPTOR stru_40E3B8;
40F178: using guessed type _UNKNOWN *off_40F178;
40F2A0: using guessed type WDFFUNCTIONS WdfFunctions;
40F470: using guessed type int (__fastcall *)(_DWORD, _DWORD, _DWORD, _DWORD, _DWORD, _DWORD);
40F970: using guessed type int WdfDriverGlobals;
40FBB4: using guessed type int dword_40FBB4;
40FBB8: using guessed type char byte_40FBB8;
415AA0: using guessed type int dword_415AA0;
415AA4: using guessed type int dword_415AA4;
415AA8: using guessed type int dword_415AA8;
415AB8: using guessed type int dword_415AB8;
415ABC: using guessed type char byte_415ABC;
415ABD: using guessed type char byte_415ABD;
[...]

PR #210, opened 2025-12-08
Add IDA Home support in README

jacobprezant:
Adds mention of IDA Home as a supported alternative to IDA Pro while pointing out lack of support for `idalib` & Feeds. #155


mrexodia:
It is not supported by _me_, so I will not be merging it for now.

Issue #127, opened 2025-08-27
Add pagination or offset to disassemble_function


hedgehogform:
I get:  Error: MCP tool "disassemble_function" response (76853 tokens) 
     exceeds maximum allowed tokens (25000). Please use pagination, 
     filtering, or limit parameters to reduce the response size.

<img width="1046" height="338" alt="Image" src="https://github.com/user-attachments/assets/cc28b257-ec80-40b6-a2e3-e81ff770810e" />

mrexodia:
You should probably be using a model with a bigger context. 25k is miniscule

ShadowNineX:
> You should probably be using a model with a bigger context. 25k is miniscule

I'm using Claude Code Sonnet 4

silentEAG:
I noticed in the CC documentation that `MAX_MCP_OUTPUT_TOKENS` can be customized, ref: https://docs.anthropic.com/en/docs/claude-code/mcp#mcp-output-limits-and-warnings. Would this help resolve this issue?

ShadowNineX:
> I noticed in the CC documentation that `MAX_MCP_OUTPUT_TOKENS` can be customized, ref: https://docs.anthropic.com/en/docs/claude-code/mcp#mcp-output-limits-and-warnings. Would this help resolve this issue?

Ah thanks! I have set that now and I'll see if I get an issue again.

NeKroFR:
I think we can close this issue @mrexodia

PR #215, opened 2025-12-13
feat: generic output size limiting middleware

### LINKED ISSUE #182: Restore data truncation function
https://github.com/mrexodia/ida-pro-mcp/blob/02e0457accd6d48697a00fe87547a259ecc26a45/src/ida_pro_mcp/ida_mcp/mcp.py#L366-L368

The idea was to save the output to a JSON file in a temporary folder and return `Function too large, read /tmp/ida-pro-mcp-outputXYZ.json instead`. This has two problems:

1. It does not respect the schema, so clients will fail
2. There is no way to retrieve a file when using the MCP remotely

Needs some thought, relates to #127

DepsCian:
# Generic Output Size Limiting for All MCP Tools

## Problem

LLMs have limited context windows (~200-400k tokens). Some MCP tools can return millions of characters in a single call, causing context overflow.

Previous approach required adding parameters to each tool individually — rejected as "case-by-case instead of generic".

## Solution

Middleware that intercepts ALL tool responses and automatically:
1. Detects large outputs (>50k chars)
2. Creates preview with truncated arrays (10 items) and strings (1000 chars)
3. Caches full output on server
4. Provides download URL for AI to retrieve full data via curl

**Generic solution — no changes needed to individual tools.**

## How It Works

```mermaid
flowchart LR
    A[MCP Tool] --> B[Middleware]
    B --> C{Size > 50k?}
    C -->|No| D[Client]
    C -->|Yes| E[Cache Full Output]
    E --> F[Preview + Download URL]
    F --> D
    G[HTTP /output/id.json] --> H[Full Output]
```

### Example Response

```json
[
  {
    "data": [
      {"addr": "0x1000", "name": "main"},
      {"addr": "0x2000", "name": "init"},
      {"_truncated": "... and 4990 more items"}
    ],
    "_output_truncated": true,
    "_total_chars": 204791,
    "_output_id": "abc-123-uuid",
    "_download_url": "http://127.0.0.1:13337/output/abc-123-uuid.json",
    "_download_hint": "Output truncated. Run: curl -o .ida-mcp/abc-123-uuid.json http://..."
  }
]
```

AI sees hint → runs curl → downloads full output to `.ida-mcp/` → uses IDE search/grep.

### rpc.py
- `_patched_mcp_tools_call` middleware patching `MCP_SERVER.registry.methods["tools/call"]`
- `_truncate_value()` — recursive preview generator (10 items, 1000 chars)
- `_add_download_info()` — adds download metadata to response
- `set_download_base_url()` for dynamic URL configuration
- Fallback to `IDA_MCP_URL` environment variable
- In-memory cache `_output_cache` for full outputs

### http.py
- `/output/{output_id}.{extension}` endpoint
- Supports .json, .c, .asm, .txt extensions
- `get_cached_output()` retrieves from cache

### ida_mcp.py & idalib_server.py
- Call `set_download_base_url()` on server start with actual host:port

## Remote MCP Support

Download URL is dynamically set based on server address:

| Scenario | Download URL |
|----------|--------------|
| Local IDA | `http://127.0.0.1:13337/output/{id}.json` |
| Remote IDA | `http://<server-ip>:<port>/output/{id}.json` |
| Custom | `$IDA_MCP_URL/output/{id}.json` |

Works for remote MCP — same address client uses to connect.

## Configuration

| Constant | Default | Description |
|----------|---------|-------------|
| `OUTPUT_LIMIT_MAX_CHARS` | 50000 | Threshold to trigger truncation |
| `OUTPUT_LIMIT_PREVIEW_ITEMS` | 10 | Max array items in preview |
| `OUTPUT_LIMIT_PREVIEW_STR_LEN` | 1000 | Max string length in preview |

## Benefits

- All 71 tools covered automatically
- Schema always valid (preserves original structure)
- Works for remote MCP
- Zero tool modifications needed
- New tools automatically covered
- AI can download full output and use native IDE search

## Related Issues

- Closes #182
- Addresses #127

### DIFF
--- a/src/ida_pro_mcp/ida_mcp.py
+++ b/src/ida_pro_mcp/ida_mcp.py
@@ -64,6 +64,11 @@ class MCP(idaapi.plugin_t):
                 MCP_SERVER.serve(
                     self.HOST, port, request_handler=IdaMcpHttpRequestHandler
                 )
+                if TYPE_CHECKING:
+                    from .ida_mcp.rpc import set_download_base_url
+                else:
+                    from ida_mcp.rpc import set_download_base_url
+                set_download_base_url(f"http://{self.HOST}:{port}")
                 print(f"  Config: http://{self.HOST}:{port}/config.html")
                 self.mcp = MCP_SERVER
                 break
--- a/src/ida_pro_mcp/ida_mcp/http.py
+++ b/src/ida_pro_mcp/ida_mcp/http.py
@@ -1,12 +1,13 @@
 import html
 import json
+import re
 import ida_netnode
 from urllib.parse import urlparse, parse_qs
 from typing import TypeVar, cast
 from http.server import HTTPServer
 
 from .sync import idaread, idawrite
-from .rpc import McpRpcRegistry, McpHttpRequestHandler, MCP_SERVER, MCP_UNSAFE
+from .rpc import McpRpcRegistry, McpHttpRequestHandler, MCP_SERVER, MCP_UNSAFE, get_cached_output
 
 
 T = TypeVar("T")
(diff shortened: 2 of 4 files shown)


mrexodia:
Cheers!

po0p:
While idea is nice, it doesn't work with claude code since it seems like the '_meta' fields are never passed to actual model (see https://github.com/anthropics/claude-code/issues/32516 )

PR #228, opened 2026-01-06
fix(jsonrpc): Better union type mismatch output & workaround MCP Clients submitting str instead of object

### LINKED ISSUE #226: LLMs pass JSON-encoded strings instead of objects, causing confusing "union does not contain str" errors
When using the `set_type` tool (and other tools with `TypedDict | list[TypedDict]` parameters), LLMs often
  incorrectly JSON-encode the parameter value. Instead of passing a native JSON object/array, they pass a JSON
  string, which causes a confusing validation error.

**LLM calls (WRONG):**
```json
  {
    "name": "set_type",
    "arguments": {
      "edits": "[{\"addr\": \"352864\", \"kind\": \"data\", \"name\": \"some_func\", \"ty\": \"void *\"}]"
    }
  }
```

  Expected call (CORRECT):
```json
  {
    "name": "set_type",
    "arguments": {
      "edits": [{"addr": "352864", "kind": "data", "name": "ut_ine_context_head", "ty": "void *"}]
    }
  }
```

The error output "Invalid params: edits union does not contain str" confuses LLM, causing him to try harder to quote the JSON string.

---

I propose some fixes here:

## Better error output (so that LLM knows what it should pass)

Around src/ida_pro_mcp/ida_mcp/zeromcp/jsonrpc.py:300

```python
if not type_matched:
      # Special hint for common LLM mistake of passing JSON strings
      if isinstance(value, str):
          # Check if any union member expects dict/list (TypedDict or list)
          expects_object = any(
              is_typeddict(arg_type) or
              (get_origin(arg_type) in (list, dict) or arg_type in (list, dict))
              for arg_type in args
              if arg_type is not type(None)
          )
          if expects_object:
              # Truncate very long strings for readability
              preview = value[:100] + "..." if len(value) > 100 else value
              raise JsonRpcException(-32602,
                  f"Invalid params: {param_name} received a JSON string but expects an object or array. "
                  f"Pass the decoded value directly, not a JSON-encoded string. "
                  f"Received string: {preview}"
              )
      raise JsonRpcException(-32602, f"Invalid params: {param_name} union does not contain {type(value).__name__}, it should contain one of {[t.__name__ if isinstance(t, type) else str(t) for t in args]}")
  validated_params[param_name] = value
  continue
```

## Try auto-decode JSON strings

Around src/ida_pro_mcp/ida_mcp/zeromcp/jsonrpc.py:290

```python
# Handle Union types (int | str, Optional[int], etc.)
  if origin in (Union, UnionType):
      type_matched = False
      for arg_type in args:
          if arg_type is type(None):
              continue

          arg_origin = get_origin(arg_type)
          check_type = arg_origin if arg_origin is not None else arg_type

          # TypedDict cannot be used with isinstance - check for dict instead
          if is_typeddict(arg_type):
              check_type = dict

          if isinstance(value, check_type):
              type_matched = True
              break

      # NEW: Try JSON decoding if string didn't match and union expects dict/list
      if not type_matched and isinstance(value, str):
          try:
              import json
              decoded = json.loads(value)
              # Check if decoded value matches any union member
              for arg_type in args:
                  if arg_type is type(None):
                      continue

                  arg_origin = get_origin(arg_type)
                  check_type = arg_origin if arg_origin is not None else arg_type

                  if is_typeddict(arg_type):
                      check_type = dict

                  if isinstance(decoded, check_type):
                      validated_params[param_name] = decoded
                      type_matched = True
                      break
          except (json.JSONDecodeError, ValueError, TypeError):
              # Not valid JSON, fall through to error below
              pass

      if not type_matched:
          # Special hint for common LLM mistake of passing JSON strings
          if isinstance(value, str):
              # Check if any union member expects dict/list (TypedDict or list)
              expects_object = any(
                  is_typeddict(arg_type) or
                  (get_origin(arg_type) in (list, dict) or arg_type in (list, dict))
                  for arg_type in args
                  if arg_type is not type(None)
              )
              if expects_object:
                  # Truncate very long strings for readability
                  preview = value[:100] + "..." if len(value) > 100 else value
                  raise JsonRpcException(-32602,
                      f"Invalid params: {param_name} received a JSON string but expects an object or array. "
                      f"Pass the decoded value directly, not a JSON-encoded string. "
                      f"Received string: {preview}"
                  )
          raise JsonRpcException(-32602, f"Invalid params: {param_name} union does not contain 
  {type(value).__name__}")
      validated_params[param_name] = value
      continue
```

## Update the set_type function arg to add annotations

```python
  def set_type(
      edits: Annotated[
          list[TypeEdit] | TypeEdit, 
          "Type edits to apply. Pass as JSON object(s), NOT as a JSON-encoded string. "
          "Example: {\"addr\": \"0x401000\", \"ty\": \"int\"} or [{\"addr\": \"0x401000\", \"ty\": \"int\"}, {...}]"
      ]
  ) -> list[dict]:
```

---

If any of above is approachable, I may submit a PR. The code above are generated by LLM, I am not fully sure that they are correct.

84634E1A607A:
I ended up still adding a generic workaround, to try to convert string to JSON object when checking type. This is because:

- Many RPC endpoints use union type, fixing each and every of them is actually more work than expected
- Some of them already have `| str`
  https://github.com/mrexodia/ida-pro-mcp/blob/8d407ca9b80e75bbba5430e5e0e297241134fbea/src/ida_pro_mcp/ida_mcp/api_core.py#L205-L214
  some don't but call `normalize_dict_list`
  https://github.com/mrexodia/ida-pro-mcp/blob/8d407ca9b80e75bbba5430e5e0e297241134fbea/src/ida_pro_mcp/ida_mcp/api_types.py#L66-L76
  some don't even call the utility function
  https://github.com/mrexodia/ida-pro-mcp/blob/8d407ca9b80e75bbba5430e5e0e297241134fbea/src/ida_pro_mcp/ida_mcp/api_analysis.py#L360-L363
  Adding `| str` and calling `normalize_dict_list` won't prevent future endpoints from this type of error
- Adding that causes linter errors (`normalize_dict_list` returns `list[dict[Unknown]]`)

Still, I reworded the union type error to be more specific and informational.

Fixes #226

### DIFF
--- a/src/ida_pro_mcp/ida_mcp/zeromcp/jsonrpc.py
+++ b/src/ida_pro_mcp/ida_mcp/zeromcp/jsonrpc.py
@@ -283,11 +283,19 @@ class JsonRpcRegistry:
                 if origin in (Union, UnionType):
                     type_matched = False
 
-                    # Try to parse str as JSON for non-str unions
+                    # HACK: Try to parse str as JSON for non-str unions
+                    # 
+                    # When JSON schema says one field is "object", Claude Code
+                    # (and maybe other MCP clients) can't (or won't) detect
+                    # that the field is actually a dict/list. Instead, they
+                    # treat the field as a string containing JSON object.
+                    #
+                    # To work around this, if the expected type is a Union
+                    # that does not include str, and the provided value is
+                    # a str, we try to parse it as JSON first.
                     if type(str) not in args and isinstance(value, str):
                         try:
                             value = json.loads(value)
-                            print(f"[MCP] Note: parsed param {param_name} from string as JSON")
                         except json.JSONDecodeError:
                             pass
 


src/ida_pro_mcp/ida_mcp/zeromcp/jsonrpc.py
```diff
@@ -282,6 +282,15 @@ def _call(self, method: str, params: Any) -> Any:
                 # Handle Union types (int | str, Optional[int], etc.)
                 if origin in (Union, UnionType):
                     type_matched = False
+
+                    # Try to parse str as JSON for non-str unions
+                    if type(str) not in args and isinstance(value, str):
+                        try:
+                            value = json.loads(value)
+                            print(f"[MCP] Note: parsed param {param_name} from string as JSON")
```
mrexodia: print is not acceptable in this code path

mrexodia:
Awesome, thanks a lot!

PR #230, opened 2026-01-10
docs: add sk installation instructions

alizain:
`sk` is a universal package manager (works across all coding agents), handles updates, etc.. (npm/cargo for agents).

would you be open to adding this?

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,17 @@ https://github.com/user-attachments/assets/65ed3373-a187-4dd5-a807-425dca1d8ee9
 
 _Note_: You need to load a binary in IDA before the plugin menu will show up.
 
+### Skills (Optional)
+
+This repository also includes an `idapython` skill for AI coding assistants. If you're using Claude Code, Codex, OpenCode, or another compatible agent, you can install it via [sk](https://github.com/803/skills-supply):
+
+```sh
+sk pkg add gh mrexodia/ida-pro-mcp --path skills
+sk sync
+```
+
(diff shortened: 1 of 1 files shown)


mrexodia:
no

alizain:
why? curious to hear your thoughts

mrexodia:
i don't want advertisements on my repo?

alizain:
I understand that, it’s certainly one way of looking at it.

I see it differently. I proposed this PR because it solves a problem for me, and I do think it solves a for you for other users of this project (otherwise, I wouldn’t have bothered). It’s annoying to use the same skills with multiple agents and keep them in updated (either globally in user-scope or for a specific project). It’s also tricky to distribute skills I or you have written, for multiple agents. This solves that.

Now if you feel it doesn’t solve it or you don’t care about the problem, I understand. I’d love some feedback for `sk`. Right now it just supports skills but I want to add support for commands and agent-specific things. It would be awesome to have a common container spec for agent extensions/packages/plugins, with some way to override for agent-specific tuning. That’s at least the direction I’m taking.

Also, stating the obvious here, but it’s not like you’re gonna get charged $0.01 every time someone uses `sk` 😅.

Anyways, thanks for sharing.

Ps. It’s a cool project you’ve made. I’ve been trying to reverse engineer Quicken on my Mac with Claude/Codex and lldb/ghidra. Trying to stop wasting time categorizing expenses when an AI can do it. If I had paid for IDA, I would have loved to use this.

PR #233, opened 2026-01-12
feat: add get_output_log tool to read IDA Output window logs

fuqiuluo:
## Summary

Add a new MCP tool `get_output_log` that retrieves messages from IDA's Output window using `ida_kernwin.msg_get_lines()`.

## Changes

- Add `api_ui.py` module for UI-related API functions
- Register the new module in `__init__.py`

## Usage

```python
# Get all log lines
get_output_log()

# Get last N lines
get_output_log(lines=100)

### DIFF
--- a/src/ida_pro_mcp/ida_mcp/__init__.py
+++ b/src/ida_pro_mcp/ida_mcp/__init__.py
@@ -26,6 +26,7 @@ from . import api_stack
 from . import api_debug
 from . import api_python
 from . import api_resources
+from . import api_ui
 
 # Re-export key components for external use
 from .sync import idasync, IDAError, IDASyncError, CancelledError
@@ -48,6 +49,7 @@ __all__ = [
     "api_debug",
     "api_python",
     "api_resources",
+    "api_ui",
     # Re-exported components
     "idasync",
     "IDAError",
--- a/src/ida_pro_mcp/ida_mcp/api_ui.py
+++ b/src/ida_pro_mcp/ida_mcp/api_ui.py
@@ -1,0 +1,21 @@
+"""UI API Functions - IDA UI interactions"""
+
+from typing import Annotated
+
+import ida_kernwin
+
+from .rpc import tool
+from .sync import idasync
+
+
+@tool
+@idasync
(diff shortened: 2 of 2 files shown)


mrexodia:
What was the use case for this?

fuqiuluo:
So I've been using the [qscripts](https://github.com/allthingsida/qscripts) plugin, which watches script files and auto-runs them whenever they change. The output shows up in IDA's Output window. Adding this `get_output_log` tool makes the workflow way better:

1. Saves tokens: Instead of constantly calling py_eval to check what happened, AI can just read the output log directly

2. Handles multi-line scripts better: py_eval has some issues with multi-line scripts, but reading from the Output window just works

3. Creates a smooth automation flow:

- qscripts watches and auto-runs script
- Output goes to the Output window
- AI grabs the results via get_output_log through MCP

mrexodia:
> Handles multi-line scripts better: py_eval has some issues with multi-line scripts, but reading from the Output window just works

Probably this should be fixed instead?

> AI grabs the results via get_output_log through MCP

I think this approach would consume much more tokens, unless the output gets cleared every time? It would also have to poll for the script being finished constantly based on some `print("Finished (script version N)")` in this workflow, seems suboptimal.

fuqiuluo:
> > Handles multi-line scripts better: py_eval has some issues with multi-line scripts, but reading from the Output window just works
> 
> Probably this should be fixed instead?
> 
> > AI grabs the results via get_output_log through MCP
> 
> I think this approach would consume much more tokens, unless the output gets cleared every time? It would also have to poll for the script being finished constantly based on some `print("Finished (script version N)")` in this workflow, seems suboptimal.

In practice, AI finds that the log is too long and will call `py_ eval` tool to execute a simple script to clear the log.

I'm not familiar with how `py_eval` works.  When AI tries to execute a very long script, it often fails multiple times and requires corrections before it can run correctly.

mrexodia:
Seems like the `py_eval` issue was fixed in https://github.com/mrexodia/ida-pro-mcp/pull/240, so I think we can avoid adding yet another tool.

PR #243, opened 2026-02-02
Add configuration dialog for MCP server host and port

taida957789:
Previously host/port were hardcoded constants. Now a dialog appears when starting the server, allowing users to configure these settings.

### DIFF
--- a/src/ida_pro_mcp/ida_mcp/zeromcp/mcp.py
+++ b/src/ida_pro_mcp/ida_mcp/zeromcp/mcp.py
@@ -307,15 +307,16 @@ class McpServer:
     def prompt(self, func: Callable) -> Callable:
         return self.prompts.method(func)
 
-    def serve(self, host: str, port: int, *, background = True, request_handler = McpHttpRequestHandler):
+    def serve(self, host: str, port: int, *, background = True, threaded = True, request_handler = McpHttpRequestHandler):
         if self._running:
             print("[MCP] Server is already running")
             return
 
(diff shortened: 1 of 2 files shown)


mrexodia:
I don't want a dialog every time you start the server.

mrexodia:
Sorry I should have clarified, I never want to show a popup to any user if they start the MCP server by hand. I'm not against a configuration dialog, but it has to be a separate action.

src/ida_pro_mcp/ida_mcp.py
```diff
@@ -43,9 +63,27 @@ def init(self):
...
         self.mcp: "ida_mcp.rpc.McpServer | None" = None
+        self.host = MCP.DEFAULT_HOST
+        self.port = MCP.DEFAULT_PORT
         return idaapi.PLUGIN_KEEP
 
     def run(self, arg):
+        # Skip dialog in batch/terminal mode, use defaults
+        if not idaapi.cvar.batch:
+            # Show configuration dialog
+            form = MCPConfigForm(self.host, self.port)
```
mrexodia: This should only happen in a dedicated menu item, not in the main plugin run function.

mrexodia (changes requested):

mrexodia:
Or just have a separate configuration menu item as suggested? 😅

mrexodia:
Thanks, I cherry-picked e0d75f4681406cb0bc4b0285e6c9ad132f99c0dc with that fix!

PR #245, opened 2026-02-03
Fix batch mode synchronization and ARM64 callees support

Dispa1r:
This PR fixes two critical issues:

## 1. Fix batch mode synchronization in _sync_wrapper

**Problem**: Batch mode was only applied in `sync_wrapper`'s `_run_with_batch` helper, which could be bypassed in certain code paths, leading to inconsistent behavior.

**Solution**: Move `idc.batch()` calls into `_sync_wrapper` to ensure batch mode is always enabled for all synchronized IDA operations.

**Benefits**:
- All operations executed via `_sync_wrapper` run in batch mode
- Consistent behavior across all synchronization paths
- Proper cleanup of batch state even on exceptions

**Changed files**: `src/ida_pro_mcp/ida_mcp/sync.py`

## 2. Fix callees function to support ARM64 and other architectures

**Problem**: The `callees` function used x86-specific instruction type checks (`NN_call`, `NN_callfi`, `NN_callni`), which failed on ARM64 and other architectures.

**Solution**: Replace architecture-specific checks with `ida_idp.is_call_insn()`, which works across all architectures.

**Benefits**:
- Proper call instruction detection on ARM64 (bl, blr)
- Support for ARM32 (bl, blx), MIPS (jal, jalr), and other architectures
- More maintainable and future-proof code

**Changed files**: `src/ida_pro_mcp/ida_mcp/api_analysis.py`

## Testing

Both fixes have been tested and verified to work correctly:
- Batch mode synchronization tested with multiple concurrent operations
- ARM64 callees tested with real ARM64 binaries

Co-Authored-By: Claude Sonnet 4.5 <noreply@anthropic.com>

### DIFF
--- a/src/ida_pro_mcp/ida_mcp/sync.py
+++ b/src/ida_pro_mcp/ida_mcp/sync.py
@@ -65,11 +65,14 @@ def _sync_wrapper(ff):
             raise IDASyncError(error_str)
 
         call_stack.put((ff.__name__))
+        # Enable batch mode for all synchronized operations
+        old_batch = idc.batch(1)
         try:
             res_container.put(ff())
         except Exception as x:
             res_container.put(x)
         finally:
+            idc.batch(old_batch)
             call_stack.get()
 
     idaapi.execute_sync(runned, idaapi.MFF_WRITE)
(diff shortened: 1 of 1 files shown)


mrexodia:
Thanks!

PR #249, opened 2026-02-07
Gate decompiler calls for safety

strazzere:
The MCP will hang if/when these are called and the decompiler is not available.

e.g. - if you try to rename things without ever using the decompiler (not a plugin that is available) it will call a refresh which then hangs the mcp.

### DIFF
--- a/src/ida_pro_mcp/ida_mcp/utils.py
+++ b/src/ida_pro_mcp/ida_mcp/utils.py
@@ -735,6 +735,8 @@ def pattern_filter(data: list[T], pattern: str, key: str) -> list[T]:
 
 
 def refresh_decompiler_widget():
+    if not ida_hexrays.init_hexrays_plugin():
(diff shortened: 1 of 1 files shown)


mrexodia:
Why does it hang when calling those functions?

strazzere:
No clue -- it seems to try and interact with the decompiler which doesn't exist and it never returns. So the function will never return.

For example -- I was running renames, the first rename "works" but never returns (it's waiting for the decompiler to refresh) so every subsequent rename seemingly is blocked as well.

Gating this function allowed me to continue using all functionality as it prevented any decompiler usage if it didn't exist/initialize.

strazzere:
To be clear when I say "is not available" - I mean I don't have the decompiler plugin - so it is expected to never succeed when using it.

mrexodia:
Alright, thanks!

PR #250, opened 2026-02-07
Fix “find_bytes” and “find” return empty results on IDA Pro 9.0

hzqst:
## Problem

The `find_bytes` tool in `ida_mcp/api_analysis.py` always returned 0 matches when searching for function byte patterns (e.g. `48 8B ?? ??`) in IDA Pro 9.0, regardless of the binary being analyzed.

### 1. `ida_bytes.bin_search` does not exist in IDA 9.0

The original code (and the first fix attempt) called `ida_bytes.bin_search()`, which is **not available** in IDA Pro 9.0. The recommended IDA 9.0 API for binary search is:

```python
ida_bytes.find_bytes(
    bs: bytes | bytearray | str,
    range_start: int,
    range_size: int | None = None,
    range_end: int | None = BADADDR,
    mask: bytes | bytearray | None = None,
    flags: int | None = BIN_SEARCH_FORWARD | BIN_SEARCH_NOSHOW,
    radix: int | None = 16,
    strlit_encoding: int | str | None = PBSENC_DEF1BPU
) -> int
```

This function accepts:
- A **string** pattern like `"48 8B ? ?"` (with `?` as byte wildcard), or
- Raw **bytes** with a **mask** for exact control.

### 2. `parse_binpat_str` return-value check was unreliable

The original code used `ida_bytes.parse_binpat_str()` + `compiled_binpat_vec_t` and checked the return value with `if err:`, which had ambiguous truth semantics across IDA versions (see prior analysis).

### 3. `except Exception: pass` silently swallowed all errors

The bare `except Exception: pass` hid the `AttributeError: module 'ida_bytes' has no attribute 'bin_search'`, making the root cause invisible to the caller. The result was always 0 matches with no error message.

### 4. Same issue in the `find` tool

The `find` function's `string` search and `immediate` search also called `ida_bytes.bin_search()` directly, making them equally broken on IDA 9.0.

### New helper: `_raw_bin_search()`

Added a version-compatible helper that abstracts binary search across IDA versions:

```python
def _raw_bin_search(ea, max_ea, data, mask, flags=0) -> int:
```

Resolution order:
1. **`ida_bytes.find_bytes()`** — IDA 9.0+ high-level API (bytes + mask)
2. **`ida_bytes.bin_search()`** — Older IDA low-level 6-param API
3. **`RuntimeError`** — If neither API is available

### `find_bytes` tool

Completely rewired the search logic:
- Removed `compiled_binpat_vec_t` and `parse_binpat_str` entirely.
- Primary path (IDA 9.0+): Normalizes `??` to `?` in the pattern string, then calls `ida_bytes.find_bytes(pattern_str, ...)` directly.
- Fallback path (older IDA): Parses to raw `bytes` + `mask`, calls `ida_bytes.bin_search()`.
- Fixed exception handling: `except Exception as e:` now propagates error messages in the result dict.

### `find` tool — `string` and `immediate` search types

Replaced all direct `ida_bytes.bin_search()` calls with `_raw_bin_search()`, which provides the same version-compatibility fallback chain.

## IDA Pattern Format Note

IDA's native pattern format uses a **single `?`** per wildcard byte (e.g. `"48 8B ? ? 90"`), while many reverse engineering tools use `??`. The `find_bytes` tool now normalizes `??` to `?` before passing to `ida_bytes.find_bytes()`.

## Files Modified

- `ida_mcp/api_analysis.py`:
  - Added `_raw_bin_search()` helper (after line 43)
  - Rewrote `find_bytes` tool search logic
  - Updated `find` tool's `string` search to use `_raw_bin_search()`
  - Updated `find` tool's `immediate` search to use `_raw_bin_search()`

## Before / After fix comparison

Before fix:

image

After fix:

image

### DIFF
--- a/src/ida_pro_mcp/ida_mcp/api_analysis.py
+++ b/src/ida_pro_mcp/ida_mcp/api_analysis.py
@@ -41,20 +41,30 @@ from .utils import (
 _IMM_SCAN_BACK_MAX = 15
 
 
-def _raw_bin_search(ea: int, max_ea: int, data: bytes, mask: bytes, flags: int = 0) -> int:
+def _raw_bin_search(
+    ea: int, max_ea: int, data: bytes, mask: bytes, flags: int = 0
+) -> int:
     """Search for raw bytes with mask, compatible across IDA versions.
 
     Returns the match address, or idaapi.BADADDR if not found.
     """
-    if hasattr(ida_bytes, 'find_bytes'):
+    if hasattr(ida_bytes, "find_bytes"):
         # IDA 9.0+: high-level API accepting bytes + mask directly
-        search_flags = flags or (ida_bytes.BIN_SEARCH_FORWARD | ida_bytes.BIN_SEARCH_NOSHOW)
-        return ida_bytes.find_bytes(data, ea, range_end=max_ea, mask=mask, flags=search_flags)
-    if hasattr(ida_bytes, 'bin_search'):
+        search_flags = flags or (
+            ida_bytes.BIN_SEARCH_FORWARD | ida_bytes.BIN_SEARCH_NOSHOW
+        )
+        return ida_bytes.find_bytes(
+            data, ea, range_end=max_ea, mask=mask, flags=search_flags
+        )
+    if hasattr(ida_bytes, "bin_search"):
         # Older IDA: low-level 6-param API
-        search_flags = flags or (ida_bytes.BIN_SEARCH_FORWARD | ida_bytes.BIN_SEARCH_NOSHOW)
+        search_flags = flags or (
+            ida_bytes.BIN_SEARCH_FORWARD | ida_bytes.BIN_SEARCH_NOSHOW
+        )
         return ida_bytes.bin_search(ea, max_ea, data, mask, len(data), search_flags)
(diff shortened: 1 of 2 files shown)


src/ida_pro_mcp/ida_mcp/api_analysis.py
```diff
@@ -1,6 +1,7 @@
 from itertools import islice
 import struct
 from typing import Annotated, Optional
+import ida_hexrays
```
mrexodia: Why is this imported?

hzqst: Probably be from my previous fix attempt

src/ida_pro_mcp/ida_mcp/api_analysis.py
```diff
@@ -1149,12 +1204,8 @@ def callgraph(
         list[str] | str, "Root function addresses to start call graph traversal from"
     ],
     max_depth: Annotated[int, "Maximum depth for call graph traversal"] = 5,
-    max_nodes: Annotated[
```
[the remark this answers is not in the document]
hzqst: reformatted.

mrexodia:
Thanks!

PR #254, opened 2026-02-17
Add separate configuration menu item for host/port

idkfrancis:
## Summary

Adds an **MCP Configuration** menu item under `Edit > Plugins` that lets users set the server host and port — without interrupting the normal server startup flow.

- **Ctrl+Alt+M** starts the server immediately, no dialog shown (unchanged behavior)
- **Edit > Plugins > MCP Configuration** opens two simple prompts to set host and port
- Configured values persist for the IDA session and are used on next server start
- Defaults remain `127.0.0.1:13337`
- Cleans up the registered action in `term()`

This addresses #89 (multiple instances on different ports) and implements the approach requested in #243 by @mrexodia:

> Or just have a separate configuration menu item as suggested?

### Usage

1. Open IDA instance #1 → Ctrl+Alt+M → starts on default `127.0.0.1:13337`
2. Open IDA instance #2 → Edit > Plugins > MCP Configuration → set port to `13338` → Ctrl+Alt+M → starts on `127.0.0.1:13338`

### Changes

- Single file modified: `src/ida_pro_mcp/ida_mcp.py`
- No changes to `mcp.py`, `idalib_server.py`, or any other file
- Resolves the `# TODO: make these configurable` comment

### DIFF
--- a/src/ida_pro_mcp/ida_mcp.py
+++ b/src/ida_pro_mcp/ida_mcp.py
@@ -6,6 +6,7 @@ It loads the actual implementation from the ida_mcp package.
 
 import sys
 import idaapi
+import ida_kernwin
 from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
@@ -23,6 +24,34 @@ def unload_package(package_name: str):
         del sys.modules[mod_name]
 
 
+CONFIG_ACTION_ID = "mcp:configure"
+CONFIG_ACTION_LABEL = "MCP Configuration"
+
+
+class MCPConfigHandler(idaapi.action_handler_t):
+    def __init__(self, plugin: "MCP"):
+        idaapi.action_handler_t.__init__(self)
+        self.plugin = plugin
+
+    def activate(self, ctx):
+        host = ida_kernwin.ask_str(self.plugin.host, 0, "MCP server host:")
+        if host is None:
+            return 0
+        port = ida_kernwin.ask_long(self.plugin.port, "MCP server port (1-65535):")
+        if port is None:
+            return 0
+        if port < 1 or port > 65535:
+            print(f"[MCP] Invalid port: {port}")
+            return 0
+        self.plugin.host = host
+        self.plugin.port = port
+        print(f"[MCP] Configuration updated: {host}:{port}")
+        return 1
+
+    def update(self, ctx):
+        return idaapi.AST_ENABLE_ALWAYS
+
+
 class MCP(idaapi.plugin_t):
     flags = idaapi.PLUGIN_KEEP
     comment = "MCP Plugin"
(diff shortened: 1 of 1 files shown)


AudriusButkevicius:
This does not work, claude fixed it for me:
```
● The problem is a timing issue. During plugin_t.init(), the IDA UI menu structure isn't fully built yet, so attach_action_to_menu on line 86 silently fails. The register_action succeeds, but the menu attachment doesn't.                                                                                        
                                                                                                                                                                                                                                                                                                                    
  The fix is to defer attach_action_to_menu until the UI is ready using UI_Hooks:                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                    
● Update(ida_mcp.py)                                                                                                                                                                                                                                                                                                
  ⎿  Added 10 lines                                                                                                                                                                                                                                                                                                 
      52          return idaapi.AST_ENABLE_ALWAYS                                                                                                                                                                                                                                                                   
      53  
      54                                                                                                                                                                                                                                                                                                            
      55 +class MCPUIHooks(ida_kernwin.UI_Hooks):                                                                                                                                                                                                                                                            
      56 +    """Defers menu attachment until the UI is fully ready."""                                                                                                                                                                                                                                             
      57 +                                                                                                                                                                                                                                                                                                   
      58 +    def ready_to_run(self):                                                                                                                                                                                                                                                                        
      59 +        ida_kernwin.attach_action_to_menu(                                                                                                                                                                                                                                                         
      60 +            "Edit/Plugins/", CONFIG_ACTION_ID, idaapi.SETMENU_APP                                                                                                                                                                                                                                  
      61 +        )                                                                                                                                                                                                                                                                                          
      62 +        self.unhook()                                                                                                                                                                                                                                                                              
      63 +                                                                                                                                                                                                                                                                                                   
      64 +                                                                                                                                                                                                                                                                                                   
      65  class MCP(idaapi.plugin_t):
      66      flags = idaapi.PLUGIN_KEEP
      67      comment = "MCP Plugin"

● Now update init() to use the hook instead of calling attach_action_to_menu directly:

● Update(ida_mcp.py)                                                                                                                                                                                                                                                                                                
  ⎿  Added 3 lines, removed 3 lines
       92                  MCPConfigHandler(self),                                                                                                                                                                                                                                                                  
       93              )                                          
       94          )                                                                                                                                                                                                                                                                                                
       95 -        ida_kernwin.attach_action_to_menu(                                                                                                                                                                                                                                                        
       96 -            "Edit/Plugins/", CONFIG_ACTION_ID, idaapi.SETMENU_APP                                                                                                                                                                                                                                        
       97 -        )                                                                                                                                                                                                                                                                                         
       95 +        # Defer menu attachment until the UI is fully initialized                                                                                                                                                                                                                                 
       96 +        self._ui_hooks = MCPUIHooks()                                                                                                                                                                                                                                                             
       97 +        self._ui_hooks.hook()                                                                                                                                                                                                                                                                     
       98  
       99          return idaapi.PLUGIN_KEEP
      100  

● The problem: attach_action_to_menu("Edit/Plugins/", ...) was called in plugin_t.init(), but at that point the IDA menu bar isn't fully constructed yet, so the call silently fails.
```

AudriusButkevicius:
My claim about it not always working is still true.

mrexodia:
Yeah don't worry, I gave it all to the clanker and completely changed the implementation. If you run into any issues on `main` let me know!

Issue #261, opened 2026-02-24
Disassembly and xref “function address” fields use the query/callsite EA, not the function entry (misleading start_ea / fn.addr)


codengine:
Codex was confused by the behavior of `disasm` and `xrefs_to` in regards to the returned start_ea and fn.addr. Here is the full feedback:

>  When using ida-pro-mcp, I ran into a recurring confusion that can cause downstream tooling to incorrectly treat interior addresses as distinct callable functions.

>  ### What I observed
>
>  1. disasm(addr=...) returns asm.start_ea equal to the requested address, even when that address is inside a function.
>
>  - Example: calling disasm(0x1651C) (an interior instruction inside _farmalloc) returns asm.name="_farmalloc" but asm.start_ea="0x1651c".
>  - IDA ground truth: ida_funcs.get_func(0x1651C).start_ea == 0x164F0.
>
>  This behavior is implemented in src/ida_pro_mcp/ida_mcp/api_analysis.py:
>
>  - header_addr = start  # Use requested address, not function start
>  - returned as "start_ea": hex(header_addr)
>
>  2. xrefs_to (and other helpers that use get_function(xref.frm)) returns fn.addr equal to the xref site/callsite EA, not the containing function entry.
>
>  - get_function(addr) returns Function(addr=hex(addr), ...) even though it already has the func_t and its start_ea.
>
>  This is implemented in src/ida_pro_mcp/ida_mcp/utils.py:
>
>  - return Function(addr=hex(addr), name=name, size=hex(fn.end_ea - fn.start_ea))
>
>  ### Why this is a problem
>
>  The field names (start_ea, fn.addr) strongly suggest “function entry address”, so consumers naturally treat them as stable function identities. In practice they can be interior addresses, which can lead to:
>
>  - seeding duplicate “function symbols” at interior addresses,
>  - creating edges/relationships to the wrong address,
>  - generating work items for non-entry addresses.
>
>  ### Expected / suggested fix
>
>  Either (A) clarify semantics via new fields, or (B) change existing fields to match their names:
>
>  Option A (non-breaking, preferred):
>
>  - For disasm: return both query_ea (or disasm_start_ea) and func_start_ea (and maybe func_end_ea).
>  - For xref “fn”: return both query_ea and start_ea, and make it clear which is which.
>
>  Option B (breaking but simpler):
>
>  - Make disasm.asm.start_ea be the actual function entry (func.start_ea) and use the existing top-level addr as “query/start point”.
[...]

Issue #237, opened 2026-01-21
lookup_funcs should return function start address


Trass3r:
https://github.com/mrexodia/ida-pro-mcp/blob/c2472d1c1e676f0198070e0e27f708973c7a4254/src/ida_pro_mcp/ida_mcp/utils.py#L529
```py
addr=hex(fn.start_ea)
```

To prevent function-related actions like comments ending up somewhere else.

Issue #93, opened 2025-06-21
Allow collapsing decompilation local variables


po0p:
Of course I wont be implementing anything, too busy vibing.
Some ideas after continious use since release. Setup: vscode + github copilot premium.
1. decompile_function(name) would be good. Some (even good) models can be stupid, and cant derive that sub_1234 is located at 0x1234.
2. decompile_function(bool collapse_locals = false) is also maybe an idea. Function big. A lot of locals. Context window small.

Thanks.

PR #264, opened 2026-02-27
Fix get_function/disasm returning interior address instead of function entry, and allow decompile/disasm to accept function names

NeKroFR:
Hi, two fixes in here.

#237 / #261: `get_function()` was returning the queried address as `fn.addr` even when it lands inside a function body, not at the entry. One-line fix: use `fn.start_ea`. `disasm()` had the exact same bug with `asm.start_ea`.

#93 (partial fix): `decompile()` and `disasm()` would blow up on anything non-numeric. I added a fallback to `idaapi.get_name_ea()` so you can pass `"main"` directly instead of hunting for the address. Unknown names give a "not found" instead of a parse error.

I tested on a small ELF with a few named functions, querying an interior address returns the entry, name lookup works on both.

Feel free to ask for changes if needed.

### DIFF
--- a/src/ida_pro_mcp/ida_mcp/api_analysis.py
+++ b/src/ida_pro_mcp/ida_mcp/api_analysis.py
@@ -246,7 +246,7 @@ def disasm(
         if func:
             # Function exists: disassemble function items starting from requested address
             func_name: str = ida_funcs.get_func_name(func.start_ea) or "<unnamed>"
-            header_addr = func.start_ea  # Use function entry point, not the queried address
+            header_addr = start  # Use requested address, not function start
         else:
             # No function: disassemble sequentially from start address
             func_name = "<no function>"
--- a/src/ida_pro_mcp/ida_mcp/tests/test_api_analysis.py
+++ b/src/ida_pro_mcp/ida_mcp/tests/test_api_analysis.py
@@ -71,6 +71,32 @@ def test_decompile_batch():
     assert len(result) == len(addrs)
 
 
+@test()
+def test_decompile_by_name():
+    """decompile accepts a function name string"""
+    import ida_funcs
+
+    fn_addr = get_any_function()
+    if not fn_addr:
+        return
+
+    name = ida_funcs.get_func_name(int(fn_addr, 16))
+    if not name:
+        return
+
+    result = decompile(name)
+    assert result.get("error") is None
+    assert result.get("code") is not None
+
+
+@test()
+def test_decompile_unknown_name():
+    """decompile returns error for unknown function name"""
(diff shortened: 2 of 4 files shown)


mrexodia:
> I tested on a small ELF with a few named functions, querying an interior address returns the entry, name lookup works on both.

Could you share this file and the addresses you used to test? It would be good to have proper tests for this...

NeKroFR:
Here the binary I used (compiled with  `gcc -O0 -no-pie -o tests/test.elf tests/test.c`)

```c
// gcc -O0 -no-pie -o tests/test.elf tests/test.c
#include <stdio.h>

int add(int a, int b) { return a + b; }
int sub(int a, int b) { return a - b; }
int mul(int a, int b) { return a * b; }
int div(int a, int b) { return a / b; }
int mod(int a, int b) { return a % b; }

int main(void) {
    printf("%d\n", add(10, 3));
    printf("%d\n", sub(10, 3));
    printf("%d\n", mul(10, 3));
    printf("%d\n", div(10, 3));
    printf("%d\n", mod(10, 3));
    return 0;
}
```

I don't think it is relevant to add the file, however if you want I can still add it. I didn't test edge cases but the changes are minimal so I don't think it should break.

mrexodia:
Currently there is only one test binary: https://github.com/mrexodia/ida-pro-mcp/blob/main/tests/crackme03.elf. It looks like your issues should be reproducible in there, just needs tests to wire up. Currently the testing infrastructure is rather mediocre, but the project has become too big and disorganized to yolo merge everything.

NeKroFR:
Added `tests/test_function_entry.py` which covers my changes using `crackme03.elf`. Let me know if you want me to change anything or if I can help with anything else.

NeKroFR:
@mrexodia, any suggestions or feedback?

mrexodia:
Sorry, been busy. For the tests there is a builtin testing mechanism `@test` which will be run with `uv run ida-mcp-test`. Just wanted to make sure about one thing:

The `disasm` function was meant to allow disassembling ranges. So if you pass the middle of a function it's supposed to disassemble from that address, not the start of the function. Otherwise you can never disassemble a 10k instruction function. This PR seems to do the exact opposite?

NeKroFR:
Yeah, you're totally right about the disasm range behavior. I reverted the `header_addr` change and dropped the issue #261.
I can still address it by adding an entry_ea field alongside the existing start_ea.

I also added tests with `@test`. However I have issues with idalib so I couldn't run `uv run ida-mcp-test` locally. So before merging, I think it would be a good idea to verify the test works.

mrexodia:
Thanks for your contribution!

PR #286, opened 2026-03-13
Add STDIO transport for idalib mode

Evian-Zhang:
This can be helpful if we want to use idalib locally with only one client


mrexodia:
It’s not possible because ida can print to stdout through plugins/script.

PR #293, opened 2026-03-15
Add stdio transport for Opencode

vee1e:
## Summary
This PR addresses additional incompatibilities for the Opencode client found after #282.
### Fixes
Opencode requires an explicit `{ "type": "...", "url": "..." }` structure for HTTP/SSE transports (identical to Claude/Claude Code). This PR adds Opencode to the list of clients using this format, resolving "Invalid input" errors when using HTTP/SSE transports.

Before this there were issues installing Opencode MCP with the HTTP option which I tried on a clean install and got JSON config issues. This should hopefully fix them.

### DIFF
--- a/src/ida_pro_mcp/installer.py
+++ b/src/ida_pro_mcp/installer.py
@@ -156,8 +156,7 @@ def generate_mcp_config(*, client_name: str, transport: str = "stdio"):
 
     transport_url = normalize_transport_url(transport)
     if client_name == "Opencode":
-        # Opencode requires local (stdio) transport
-        return generate_mcp_config(client_name=client_name, transport="stdio")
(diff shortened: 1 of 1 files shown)


vee1e:
sorry for the ping @mrexodia but can this merge?

mrexodia:
Not really, it seems incorrect that only stdio is supported and if that's the case opencode shouldn't show up when selecting http transports in the first place.

https://github.com/anomalyco/opencode/issues/8058

vee1e:
hmm let me see this and get back to you

vee1e:
seems the error was just a simple change in the schema, for HTTP transports Opencode uses `remote` instead of `http`.

mrexodia:
cheers!

PR #299, opened 2026-03-17
feat: Unix domain socket support, execute_sync deadlock fix, Bearer token auth

zh-explorer:
## Summary

  Three small, independent improvements to the core infrastructure. Each is backward compatible — no
  behavior change when new options are unused.

  ### 1. Unix domain socket support for idalib server

  **Commits:** `3a571b5`

  Adds `UnixHTTPServer` and `UnixThreadingHTTPServer` classes to zeromcp, and a `unix_socket` parameter to
   `McpServer.serve()`. The idalib server gains a `--unix-socket <path>` CLI flag.

  **Motivation:** When running multiple idalib instances (e.g. from a pool manager), TCP ports can
  conflict or become exhausted. Unix domain sockets are more reliable for local IPC and have lower
  overhead.

  **Technical note:** `HTTPServer.server_bind()` assumes `(host, port)` tuple from `getsockname()`, but
  AF_UNIX returns a path string. The mixin skips `HTTPServer.server_bind` and calls
  `TCPServer.server_bind` directly. Both single-threaded (`UnixHTTPServer`) and multi-threaded
  (`UnixThreadingHTTPServer`) variants are provided, matching the existing
  `HTTPServer`/`ThreadingHTTPServer` split.

  ### 2. Fix execute_sync deadlock in headless idalib mode

  **Commits:** `ef136b4`

  In headless idalib mode with `background=False`, the main thread runs `serve_forever()`. When a
  single-threaded server handles a request on the main thread, `@idasync` calls `idaapi.execute_sync()`
  which queues a callback for the main thread — the same thread that's waiting — causing a deadlock.

  **Empirically verified:** `execute_sync` from a worker thread deadlocks in idalib headless mode because
  there is no IDA event loop to dispatch queued callbacks. Direct calls on the main thread work fine.

  **Fix:** Capture the main thread ID at module load time. In `_sync_wrapper`, if already on the main
  thread, execute directly. The shared logic (batch mode + call-stack reentry detection) is extracted into
   `_run_with_batch()` so both the direct path and the `execute_sync` path use identical guards.

  ### 3. Bearer token authentication

  **Commits:** `2e5de3b`

  Adds optional `--auth-token` flag (or `IDA_MCP_AUTH_TOKEN` env var) to `server.py` and
  `idalib_server.py`. When set, all HTTP requests must include `Authorization: Bearer <token>` or receive
  401.

  Implementation: `McpServer.auth_token` attribute + `_check_auth()` in `McpHttpRequestHandler`, called at
   the top of `do_GET` and `do_POST`. `Authorization` is added to CORS allowed headers.

  ## Test plan

  - [x] `uv run ida-mcp-test tests/crackme03.elf -q` — 159 passed
  - [x] `uv run ida-mcp-test tests/typed_fixture.elf -q` — 162 passed, 1 skipped
  - [x] Unix socket: idalib_server listens on socket, HTTP client connects via AF_UNIX, tools/list returns
   66 tools
  - [x] Deadlock: empirically confirmed execute_sync deadlocks from worker thread; main-thread detection
  bypasses correctly
  - [x] Auth: no token → 401, wrong token → 401, correct token → 200, GET without token → 401

### DIFF
--- a/src/ida_pro_mcp/ida_mcp/zeromcp/mcp.py
+++ b/src/ida_pro_mcp/ida_mcp/zeromcp/mcp.py
@@ -135,9 +135,20 @@ class McpHttpRequestHandler(BaseHTTPRequestHandler):
         self.send_header("Access-Control-Allow-Origin", origin)
         if preflight:
             self.send_header("Access-Control-Allow-Methods", "POST, GET, OPTIONS")
-            self.send_header("Access-Control-Allow-Headers", "Content-Type, Accept, X-Requested-With, Mcp-Session-Id, Mcp-Protocol-Version")
+            self.send_header("Access-Control-Allow-Headers", "Content-Type, Accept, X-Requested-With, Mcp-Session-Id, Mcp-Protocol-Version, Authorization")
             if self.headers.get("Access-Control-Request-Private-Network") == "true":
                 self.send_header("Access-Control-Allow-Private-Network", "true")
+
+    def _check_auth(self) -> bool:
+        """Validate Bearer token if configured.  Returns True if OK."""
+        token = self.mcp_server.auth_token
+        if token is None:
+            return True
+        auth = self.headers.get("Authorization", "")
+        if auth != f"Bearer {token}":
+            self.send_error(401, "Unauthorized")
+            return False
+        return True
 
     def send_error(self, code, message=None, explain=None):
         self.send_response(code)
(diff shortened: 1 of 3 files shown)


mrexodia:
What MCP client supports unix sockets?

mrexodia:
The deadlock fix doesn't fix anything, it's fundamentally impossible. You can easily validate this by setting `background=True` in `src/ida_pro_mcp/idalib_server.py` (as well as adding a call to `input` for an infinite loop). The fundamental problem is that IDA doesn't run a loop so you will not be able to `execute_sync`. When on the main thread `execute_sync` already executes the callback directly.

Issue #214, opened 2025-12-12
AttributeError: 'func_t' object has no attribute 'name'


chancerly:
环境：ubuntu 22.04 ida pro 9.0
```
Traceback (most recent call last):
  File "/home/xxx/.idapro/plugins/mcp-plugin.py", line 144, in do_POST
    result = rpc_registry.dispatch(request["method"], request.get("params", []))
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/xxx/.idapro/plugins/mcp-plugin.py", line 58, in dispatch
    return func(*converted_params)
           ^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/xxx/.idapro/plugins/mcp-plugin.py", line 377, in wrapper
    return sync_wrapper(ff, idaapi.MFF_READ)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/xxx/.idapro/plugins/mcp-plugin.py", line 351, in sync_wrapper
    raise res
  File "/home/xxx/.idapro/plugins/mcp-plugin.py", line 341, in runned
    res_container.put(ff())
                      ^^^^
  File "/home/xxx/.idapro/plugins/mcp-plugin.py", line 918, in disassemble_function
    if label and label == func.name and address == func.start_ea:
                          ^^^^^^^^^
AttributeError: 'func_t' object has no attribute 'name'
```

mrexodia:
Try 9.2

Issue #125, opened 2025-08-26
disassemble_function error


wt-Yoha:
i use ida pro 9.0 when the ai agent try to use function disassemble_function it's got :

Traceback (most recent call last):
  File "C:/Users/admin/AppData/Roaming/Hex-Rays/IDA Pro/plugins/mcp-plugin.py", line 144, in do_POST
    result = rpc_registry.dispatch(request["method"], request.get("params", []))
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:/Users/admin/AppData/Roaming/Hex-Rays/IDA Pro/plugins/mcp-plugin.py", line 58, in dispatch
    return func(*converted_params)
           ^^^^^^^^^^^^^^^^^^^^^^^
  File "C:/Users/admin/AppData/Roaming/Hex-Rays/IDA Pro/plugins/mcp-plugin.py", line 377, in wrapper
    return sync_wrapper(ff, idaapi.MFF_READ)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:/Users/admin/AppData/Roaming/Hex-Rays/IDA Pro/plugins/mcp-plugin.py", line 351, in sync_wrapper
    raise res
  File "C:/Users/admin/AppData/Roaming/Hex-Rays/IDA Pro/plugins/mcp-plugin.py", line 341, in runned
    res_container.put(ff())
                      ^^^^
  File "C:/Users/admin/AppData/Roaming/Hex-Rays/IDA Pro/plugins/mcp-plugin.py", line 918, in disassemble_function
    if label and label == func.name and address == func.start_ea:
                          ^^^^^^^^^
AttributeError: 'func_t' object has no attribute 'name'

mrexodia:
Try IDA 9.1

PawiX25:
I’ve tried both IDA 9.1 and 9.2, and the problem happens on both.

TrueBers:
same on 8.3

Issue #121, opened 2025-08-18
get some error during analysis


b1gcat:
I need some help

```
  File "/Users/b1gcat/.idapro/plugins/mcp-plugin.py", line 144, in do_POST
    result = rpc_registry.dispatch(request["method"], request.get("params", []))
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/b1gcat/.idapro/plugins/mcp-plugin.py", line 58, in dispatch
    return func(*converted_params)
           ^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/b1gcat/.idapro/plugins/mcp-plugin.py", line 377, in wrapper
    return sync_wrapper(ff, idaapi.MFF_READ)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/b1gcat/.idapro/plugins/mcp-plugin.py", line 351, in sync_wrapper
    raise res
  File "/Users/b1gcat/.idapro/plugins/mcp-plugin.py", line 341, in runned
    res_container.put(ff())
                      ^^^^
  File "/Users/b1gcat/.idapro/plugins/mcp-plugin.py", line 918, in disassemble_function
    if label and label == func.name and address == func.start_ea:
                          ^^^^^^^^^
AttributeError: 'func_t' object has no attribute 'name'





Traceback (most recent call last):
  File "/Users/b1gcat/.idapro/plugins/mcp-plugin.py", line 144, in do_POST
    result = rpc_registry.dispatch(request["method"], request.get("params", []))
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/b1gcat/.idapro/plugins/mcp-plugin.py", line 58, in dispatch
    return func(*converted_params)
           ^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/b1gcat/.idapro/plugins/mcp-plugin.py", line 363, in wrapper
    return sync_wrapper(ff, idaapi.MFF_WRITE)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/b1gcat/.idapro/plugins/mcp-plugin.py", line 351, in sync_wrapper
    raise res
[...]

Issue #104, opened 2025-07-11
[BUG REPO] disassemble_function/rename_stack_frame_variable error


Lioncky:
IDA 9.0Beta
> disassemble_function 
```
[{"type": "text", "text": "Error executing tool disassemble_function: JSON-RPC error -32603: Internal error (please report a bug)\nTraceback (most recent call last):\n  File \"C:/Users/ADMIN/AppData/Roaming/Hex-Rays/IDA Pro/plugins/mcp-plugin.py\", line 144, in do_POST\n    result = rpc_registry.dispatch(request[\"method\"], request.get(\"params\", []))\n             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n  File \"C:/Users/ADMIN/AppData/Roaming/Hex-Rays/IDA Pro/plugins/mcp-plugin.py\", line 58, in dispatch\n    return func(*converted_params)\n           ^^^^^^^^^^^^^^^^^^^^^^^\n  File \"C:/Users/ADMIN/AppData/Roaming/Hex-Rays/IDA Pro/plugins/mcp-plugin.py\", line 377, in wrapper\n    return sync_wrapper(ff, idaapi.MFF_READ)\n           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n  File \"C:/Users/ADMIN/AppData/Roaming/Hex-Rays/IDA Pro/plugins/mcp-plugin.py\", line 351, in sync_wrapper\n    raise res\n  File \"C:/Users/ADMIN/AppData/Roaming/Hex-Rays/IDA Pro/plugins/mcp-plugin.py\", line 341, in runned\n    res_container.put(ff())\n                      ^^^^\n  File \"C:/Users/ADMIN/AppData/Roaming/Hex-Rays/IDA Pro/plugins/mcp-plugin.py\", line 918, in disassemble_function\n    if label and label == func.name and address == func.start_ea:\n                          ^^^^^^^^^\nAttributeError: '`func_t`' object has no attribute 'name'\n", "uuid": "33141102-77f2-4dcc-b0e4-2405598f42bb"}]
```
> rename_stack_frame_variable 
```
Error executing tool rename_stack_frame_variable: JSON-RPC error -32603: Internal error (please report a bug)
Traceback (most recent call last):
[...]

PR #305, opened 2026-03-21
Fix IDA 9.0 early build API compatibility

JohnsterID:
IDA 9.0 build 240925 removed several methods that were added in IDA 8.5 but were later reinstated in IDA 9.1+. This commit adds runtime detection using hasattr() to handle these missing methods gracefully.

API Compatibility Fixes:
- func_t.get_name() missing in early 9.0 builds
- func_t.get_prototype() missing in early 9.0 builds
- tinfo_t.get_udm() missing in early 9.0 builds

parse_decl Return Value Fixes:
- In IDA 9.0, parse_decl() returns '' (empty string) on success
- Changed checks from 'if parse_decl(...)' to 'if parse_decl(...) is not None'
- Applied to _parse_type_tinfo(), _parse_function_tinfo(), and get_type_by_name()

tinfo_get_udm Fixes:
- get_udm_by_tid() returns 0 on success (C convention)
- Changed to check if udm.name is populated instead of return value

Type Lookup Fixes:
- Added 'signed __int64' and 'unsigned __int64' to type lookup table
- These are valid type names in IDA 9.0+ that were missing

Test Updates:
- Updated test_declare_stack_invalid_type_error to check for 'NoSuchType' in error message (more robust across IDA versions)

Tested: IDA 9.0.240925 and IDA 9.3.260213 - all tests pass

Related issues: #104, #121, #125, #214

### DIFF
--- a/src/ida_pro_mcp/ida_mcp/api_modify.py
+++ b/src/ida_pro_mcp/ida_mcp/api_modify.py
@@ -9,6 +9,7 @@ import ida_dirtree
 import ida_funcs
 import ida_ua
 
+from .compat import tinfo_get_udm
 from .rpc import tool
 from .sync import idasync, IDAError
 from .utils import (
@@ -567,7 +568,7 @@ def rename(batch: RenameBatch | dict) -> dict:
                         break
                     continue
 
-                idx, udm = frame_tif.get_udm(old_name)
+                idx, udm = tinfo_get_udm(frame_tif, old_name)
                 if not udm:
                     result = {
                         "func_addr": func_addr,
--- a/src/ida_pro_mcp/ida_mcp/api_stack.py
+++ b/src/ida_pro_mcp/ida_mcp/api_stack.py
@@ -9,6 +9,7 @@ import ida_typeinf
 import ida_frame
 import idaapi
 
+from .compat import tinfo_get_udm
 from .rpc import tool
 from .sync import idasync
 from .utils import (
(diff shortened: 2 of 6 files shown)


mrexodia:
What about 9.0.241217 (SP1)?

JohnsterID:
Re-tested with 9.0.241217, 9.1.250226 and 9.2.250908 all have the APIs. Only 9.0.240925 is missing them.

mrexodia:
Okay so maybe we should just explicitly mark any `9.0` with build `<241217` as unsupported instead (crash on plugin load)?

I think 8.5 was a special version for customers with active perpetual licenses around the release of 9.0. SP1 was probably addressing some of these API incompatibilities so I don't think it makes sense to support 9.0 SP0 with additional hacks if they are fixed in SP1.

JohnsterID:
This change is small to get it to work but thanks for pointing it out, I can see the 8.5 and 9.0sp1 release notes show the new API methods. If dropping 9.0 to set these API methods as a minimum then `(8.3 or higher, 9 recommended), IDA Free is not supported` should probably become `(8.5, 9.0sp1 or higher recommended), IDA Free is not supported`?

mrexodia:
🤷‍♂️ I don’t think it’s necessary to update the README, we can just fail loading the plugin on the one bad egg.

JohnsterID:
Closed for https://github.com/mrexodia/ida-pro-mcp/pull/316

PR #319, opened 2026-03-29
Add loader args in idalib_open

gzz2000:
Agent will be able to choose architecture and other load options during binary open in idalib-mcp. More work is needed later to teach agent how to do it in annotations.


mrexodia:
What argument were you passing exactly, what is the use case?

Issue #327, opened 2026-03-31
find_regex returns undeclared n field and fails MCP structured-content validation


packmad:
Ida Pro version: 9.3

When calling `find_regex` through an MCP client that validates tool responses against the declared schema, the tool fails validation because the response includes an undeclared top-level field `n`.

## Example

Call:

```json
{"pattern":"http|https|www\\.|\\.ru|\\.com|\\.php|\\.js|startpage|user\\-agent|cookie|registry|browser|favicon|search|up..."}
```

## Actual response

```json
{
  "n": 2,
  "matches": [
    {"addr": "0x40bda4", "string": "InternetSetPerSiteCookieDecisionW"},
    {"addr": "0x40bdc8", "string": "InternetClearAllPerSiteCookieDecisions"}
  ],
  "cursor": {"done": true}
}
```

## Validation error

```text
Invalid structured content returned by tool find_regex:
Additional properties are not allowed ('n' was unexpected)
```

## Why this happens

The tool schema appears to allow only:

- `matches`
- `cursor`
- `error`

with `additionalProperties: false`, but the implementation also returns `n`.

## Expected

One of the following:

- Add `n` to the schema
- Remove `n` from the response
- Relax the schema to allow additional properties

This is harmless for some clients, but strict MCP clients surface it as a tool failure.

mrexodia:
What client are you using exactly?

packmad:
> What client are you using exactly?

`openai-agents==0.13.4`

PR #329, opened 2026-04-01
Fixed three bugs

packmad:
Note: I'm sorry, it's not good practice to include so many fixes in a single PR, but I'm swamped!

# Fix 1
#327 
The function returns `"n": len(matches)` in the result dict, but `FindRegexResult` didn't declare n, and _typed_dict_to_schema generates `additionalProperties: false`, so the extra property was rejected.

So, added n: int to the `FindRegexResult TypedDict`

# Fix 2

When  `_type_to_json_schema(Any)`  is used in a large schema conversion, I added `if py_type is Any: return {}` that ensures that `typing.Any` is translated into the JSON Schema equivalent of "no restrictions."

# Fix 3
`stack_frame` is a field in the `DisassemblyFunction TypedDict` that represents the local variables allocated on the stack frame of a disassembled function.
f there's no function at the address (e.g. disassembling raw code without function boundaries), or the function has no recognized stack variables, the field is simply omitted. Without NotRequired, the generated JSON Schema would list stack_frame in "required", and the MCP client would reject any response missing it.

### DIFF
--- a/src/ida_pro_mcp/ida_mcp/api_composite.py
+++ b/src/ida_pro_mcp/ida_mcp/api_composite.py
@@ -40,7 +40,7 @@ class BasicBlockSummary(TypedDict):
 class AnalyzeFunctionResult(TypedDict, total=False):
     addr: str
     name: str
-    prototype: str
+    prototype: str | None
     size: int
     decompiled: str | None
     decompile_truncated: int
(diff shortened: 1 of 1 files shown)


mrexodia:
Cheers!

PR #332, opened 2026-04-02
Fix linux arrow key navigation and add j/k bindings

CloverGit:


mrexodia:
Thanks, but I don’t want jk bindings.

Issue #185, opened 2025-11-18
Restore `@unsafe` functionality


mrexodia:
When using the UI we can prompt the user with a YES/NO message box when unsafe functions are executed (+do not ask again for this session). Likely there is some builtin IDA API for this dialog already, so we can avoid Qt.

For `idalib-mcp` we should add the `--unsafe` flag back and remove all the unsafe methods if not set.

mrexodia:
You can now manually enable/disable tools, but unsafe tools are still enabled per default.

mrexodia:
Restored — the `@unsafe` decorator is back (`rpc.py:154`), gated behind `--unsafe` for idalib (`idalib_server.py:504`, `idalib_supervisor.py:1500`) and a toggle in the GUI plugin. Closing as done.

PR #335, opened 2026-04-03
fix: gate unsafe tools in idalib via --unsafe flag (#185)

JohnsterID:
Partially addresses #185 (idalib `--unsafe` flag only; plugin UI dialog deferred to separate PR)

The `--unsafe` flag was parsed by argparse in `idalib_server.py` but never wired up, so `@unsafe`-decorated tools were always available regardless of the flag.

### Changes

- **`idalib_server.py`**: when `--unsafe` is not passed, remove `@unsafe` tools from `MCP_SERVER.tools.methods` before serving
- **`test_server.py`**: 5 new tests verify registration, hiding, rejection, and that safe tools are unaffected

No changes to `zeromcp/mcp.py` or `rpc.py` — the generic MCP layer stays unaware of unsafe semantics. The IDA plugin is unaffected: everything stays enabled by default per the existing per-tool config UI design.

### Verified
- Plugin (IDA 9.3 GUI): 61 tools listed, `py_eval` callable ✅
- idalib `crackme03.elf`: 252 passed ✅
- idalib `typed_fixture.elf`: 248 passed, 1 skipped ✅

### DIFF
--- a/src/ida_pro_mcp/ida_mcp/tests/test_server.py
+++ b/src/ida_pro_mcp/ida_mcp/tests/test_server.py
@@ -1,10 +1,11 @@
-"""Tests for the top-level stdio proxy server (server.py)."""
+"""Tests for the top-level stdio proxy server (server.py) and unsafe tool gating."""
 
 import contextlib
 import os
 import sys
 
 from ..framework import test
+from ..rpc import MCP_SERVER, MCP_UNSAFE
 
 try:
     from ida_pro_mcp import server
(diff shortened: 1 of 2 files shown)


mrexodia:
How does this work the IDA plugin itself? I made the choice to enable everything per default and let the users configure the tools individually.

JohnsterID:
Let me re-look at this. I missed that `unsafe_enabled = False` would probably break it.

JohnsterID:
Ok, I've only touched idalib. UI work can be looked at later.

mrexodia:
Thanks!

PR #340, opened 2026-04-05
more mcps

ntkrnl64:
also format with ruff and add ty ignore


mrexodia:
no thanks, this unreviewable because of the huge amount of changes

PR #345, opened 2026-04-06
test: add coverage for @unsafe/@ext decorator sets and extension gating

JohnsterID:
Tests exercise the real MCP_UNSAFE set, MCP_EXTENSIONS dict, and MCP_SERVER.tools.methods registry populated by actual decorator execution at import time — no mocks.

Covers gaps from PR #335 review:
- @unsafe set completeness (py_eval, py_exec_file, diff_before_after, 15+ dbg_)
- @unsafe/@safe disjointness (core tools never marked unsafe)
- @ext('dbg') group exists and all members are also @unsafe
- Extension tools hidden from tools/list when no ext enabled
- Extension tools appear when ext group is enabled
- ORIGINAL_TOOLS snapshot includes discovery tools
- ORIGINAL_TOOLS covers all plugin-side registrations


mrexodia:
🙏🏻

Issue #323, opened 2026-03-30
connot connect ida-pro-mcp to claude-code on macos26


KSroido:
python env: clean conda python 3.13
ida: valid ida pro 9.0
macos version: 26.3
install steps: strictly follow the install video. Select claude code and claude
config file:
claude:
```
{
  "mcpServers": {
    "ida-pro-mcp": {
      "type": "http",
      "url": "http://127.0.0.1:13337/mcp"
    }
  },
  "preferences": {
    "coworkWebSearchEnabled": true,
    "coworkScheduledTasksEnabled": true,
    "ccdScheduledTasksEnabled": true,
    "sidebarMode": "code"
  }
}

```
claude code 
```
  "mcpServers": {                                                                                
    "ida-pro-mcp": {                                                                             
      "type": "http",                                                                            
      "url": "http://127.0.0.1:13337/mcp"                                                        
    }                                                                                            
  }
```

logs in  IDA

```
MCP] Cached 154816 strings in 4511ms
[MCP] Server started:
  Streamable HTTP: http://127.0.0.1:13337/mcp
  SSE: http://127.0.0.1:13337/sse
  Config: http://127.0.0.1:13337/config.html
[MCP] >> initialize({"protocolVersion": "2025-11-25", "capabilities": {"roots": {}, "elicitation": {"form": {}, "url": {}}}, "clientInfo": {"name": "claude-code", "version": "2.1.87"}})
[MCP] << initialize (2.8ms) {"protocolVersion": "2025-06-18", "capabilities": {"tools": {}, "resources": {"subscribe": false, "listChanged": false}, "prompts": {}}, "serverInfo": {"name": "ida-pro-mcp", "version": "1.0.0"}}
[MCP] >> notifications/initialized(null)
[MCP] << notifications/initialized (0.0ms) ERROR: Method 'notifications/initialized' not found
[MCP] >> initialize({"protocolVersion": "2025-11-25", "capabilities": {"roots": {}, "elicitation": {"form": {}, "url": {}}}, "clientInfo": {"name": "claude-code", "version": "2.1.87"}})
[MCP] << initialize (0.0ms) {"protocolVersion": "2025-06-18", "capabilities": {"tools": {}, "resources": {"subscribe": false, "listChanged": false}, "prompts": {}}, "serverInfo": {"name": "ida-pro-mcp", "version": "1.0.0"}}
[MCP] >> notifications/initialized(null)
[...]

PR #346, opened 2026-04-06
Fix: "ERROR: Method 'notifications/initialized' not found"

hzqst:
### Related issue

#323

### Previous repro

```
[MCP] >> notifications/initialized(null)
[MCP] << notifications/initialized (0.0ms) ERROR: Method 'notifications/initialized' not found
```

### Fix

Register notifications/initialized handler so the standard MCP handshake notification doesn't error


mrexodia:
Thanks!

PR #347, opened 2026-04-06
Feature/sigmaker support

rm-NoobInCoding:
Vendors the core engine from [sigmaker.py](https://github.com/mahmoudimus/ida-sigmaker) as `_sigmaker.py`
Adds `api_sigmaker.py` with 5 new MCP tools:
- `make_signature` — generate the shortest unique byte signature for an address
- `make_signature_for_function` — generate a unique signature at a function's entry point
- `make_signature_for_range` — encode a specific address range as a signature (e.g. a selection)
- `find_xref_signatures` — find signatures for code that references a target address (ideal for data/strings/vtables)
- `scan_signature` — scan the IDB for all matches of a signature pattern

All tools support batch input, name resolution, and 4 output formats: IDA (`48 8B ? EC`), x64dbg (`48 8B ?? EC`),
mask (`\x48\x8B\x00\xEC xx?x`), bitmask (`0x48, 0x8B, 0x00, 0xEC 0b1101`)
Multi-architecture wildcard policies: x86, ARM, MIPS, PPC

### DIFF
--- a/src/ida_pro_mcp/ida_mcp/api_sigmaker.py
+++ b/src/ida_pro_mcp/ida_mcp/api_sigmaker.py
@@ -1,0 +1,409 @@
+"""Signature creation and scanning tools for IDA Pro MCP.
+
+This module integrates sigmaker.py functionality to provide:
+- Unique signature generation for addresses/functions
+- Range-based signature generation (selection)
+- XREF-based signature discovery
+- Signature scanning/searching
+- Multiple output formats: IDA, x64dbg, mask, bitmask
+"""
+
+from typing import Annotated, NotRequired, TypedDict
+
+import idaapi
+import ida_funcs
+
+from .rpc import tool
+from .sync import idasync
+from .utils import parse_address, normalize_list_input
+
+# ---------------------------------------------------------------------------
+# Lazy import of sigmaker (lives in the IDA plugins dir or sys.path)
+# ---------------------------------------------------------------------------
+_sigmaker = None
+
+
+def _get_sigmaker():
+    global _sigmaker
+    if _sigmaker is None:
+        import sigmaker as _sm
+
+        _sigmaker = _sm
+    return _sigmaker
+
+
+# ---------------------------------------------------------------------------
+# Output format helpers
+# ---------------------------------------------------------------------------
+
+_FORMAT_ALIASES = {
+    "ida": "ida",
+    "x64dbg": "x64dbg",
+    "mask": "mask",
+    "bitmask": "bitmask",
+}
+
+
+def _resolve_format(fmt: str) -> "str":
+    key = fmt.lower().strip()
+    if key not in _FORMAT_ALIASES:
+        raise ValueError(
+            f"Unknown signature format '{fmt}'. "
+            f"Valid formats: ida, x64dbg, mask, bitmask"
+        )
+    return _FORMAT_ALIASES[key]
+
+
+def _make_config(
+    fmt: str,
+    wildcard_operands: bool = True,
+    continue_outside_function: bool = True,
+    max_length: int = 1000,
+) -> "object":
+    sm = _get_sigmaker()
+    return sm.SigMakerConfig(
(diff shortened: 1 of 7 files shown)


rm-NoobInCoding:
All tests passed. PR is ready for review.

mrexodia:
Out of scope.

mrexodia:
I will consider it actually

rm-NoobInCoding:
Thanks for considering it.

Just to give some context on why I think this fits into the scope of this MCP: I use this MCP heavily for writing injected DLLs and function hooks. The main issue is that LLMs are notoriously terrible at generating byte signatures directly from disassembly due to how they tokenize text. They constantly hallucinate opcodes or hardcode absolute addresses, which means the generated hooks silently fail.

src/ida_pro_mcp/ida_mcp/api_sigmaker.py
```diff
@@ -0,0 +1,397 @@
...
+                "addr": hex(ea) if ea is not None else None,
+                "signatures": None,
+                "error": str(e),
+            })
+    return results
+
+
+@tool
+@idasync
+def scan_signature(
```
mrexodia: Doesn't this already exist?

mrexodia (changes requested):
The tests definitely need work to be semantically meaningful vs spinning cycles not validating anything meaningful.

rm-NoobInCoding:
Addressed all review comments. Here's what each test now validates:

 - `test_make_signature_round_trip` — Generates a signature for a function address, then scans it back using find_bytes to
verify it returns exactly 1 match at the original address. This proves the signature is genuinely unique rather than just trusting the unique: True flag.

 - `test_make_signature_invalid_address` — Passes an unmapped address and asserts the result contains an error. Tests the error-handling path.

 - `test_make_signature_batch` — Passes multiple hex addresses in a single call and verifies each result maps back to the correct query address with a non-null signature. Tests the batch/list input path.

 - `test_make_signature_by_name` — Passes a known function name (check_pw) instead of a hex address and verifies the signature is generated and unique. Tests name resolution in make_signature.

 - `test_make_signature_for_function_resolves_to_start` — Passes a mid-function address (start + 1) and verifies the result addr resolves to the function start, and that the name field matches idaapi.get_func_name(). This is the key behavior that distinguishes make_signature_for_function from make_signature.

 - `test_make_signature_for_function_by_name` — Passes "main" by name and verifies both query and name fields are correctly populated. Tests name-to-function resolution.

 - `test_make_signature_for_function_no_func` — Passes a data segment address (no function) and asserts the result contains an error. Tests the "not a function" error path.

 - `test_make_signature_for_function_batch_mixed_input` — Passes a mix of hex addresses and symbolic names in a single batch, verifies each result has a correct name field, and that they resolve to different addresses. This exercises the name-resolution path that a pure-hex batch doesn't cover.

 - `test_make_signature_for_range_valid` — Takes a real function's first 16 bytes as a range and verifies a non-empty signature is produced with a valid address.

 - `test_make_signature_for_range_crackme` — Uses a known address range from crackme03.elf to verify range-based signature generation on a specific binary.

 - `test_find_xref_signatures_for_string` — Uses a known string address (0x2004) in crackme03.elf and verifies that xref signatures are found with valid signature strings, lengths, and xref addresses.

 - `test_find_xref_signatures_no_xrefs` — Passes an unmapped address and verifies either an error or zero xrefs are reported.

 - `test_all_output_formats` — Generates signatures in all 4 formats (ida, x64dbg, mask, bitmask) and validates format-specific characteristics: ?? in x64dbg, \x in mask, 0b in bitmask.

 - `test_name_resolution_across_tools` — Passes the same function name to both make_signature and make_signature_for_function and verifies they resolve to the same address. Tests consistency of name resolution across tools.

bckd00r:
when this is approved and in production?

mrexodia:
Whenever I have time to review properly. Likely ~3 weeks from now.

mrexodia:
Thanks! I merged it manually

PR #352, opened 2026-04-07
fix: validate Host and Origin on MCP HTTP endpoints

ZX41R:
## Summary
- validate `Host` on loopback-bound MCP HTTP endpoints instead of relying on CORS alone
- reject browser requests with disallowed `Origin` values while still allowing direct clients with no `Origin`
- apply the same guard to `/output/*` downloads and add regression coverage for the transport boundary

## Rationale
The config UI already validates browser-facing requests, but the shared MCP HTTP transport still accepted browser requests as long as they reached the loopback listener. Same-origin DNS rebinding can bypass passive CORS handling, so the request handler should actively reject unexpected `Host` and `Origin` headers on `/mcp`, `/sse`, and `/output/*`.

## Testing
- `python -m unittest discover -s tests -p 'test_browser_transport_guards.py'`
- `PYTHONPATH=src python tests/test_server_transport.py`
- `python -m py_compile src/ida_pro_mcp/ida_mcp/zeromcp/mcp.py src/ida_pro_mcp/ida_mcp/http.py`

### DIFF
--- a/src/ida_pro_mcp/ida_mcp/http.py
+++ b/src/ida_pro_mcp/ida_mcp/http.py
@@ -138,6 +138,8 @@ class IdaMcpHttpRequestHandler(McpHttpRequestHandler):
         # Handle output download requests
         output_match = re.match(r"^/output/([a-f0-9-]+)\.(\w+)$", path)
         if output_match:
+            if not self._check_api_request():
+                return
             self._handle_output_download(output_match.group(1), output_match.group(2))
             return
 
--- a/src/ida_pro_mcp/ida_mcp/zeromcp/mcp.py
+++ b/src/ida_pro_mcp/ida_mcp/zeromcp/mcp.py
@@ -7,6 +7,7 @@ import uuid
 import json
 import gzip
 import zlib
+import ipaddress
 import inspect
 import threading
 import traceback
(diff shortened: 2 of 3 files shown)


mrexodia:
I need to see a real reproduction of this 'vulnerability' before fixing it.

ZX41R:
I reproduced this in a real browser on Windows rather than via curl-only reasoning.

Setup:
- Victim service: IDA MCP bound to `127.0.0.1:13337` on the Windows host (`ida.exe` was the only listener on that port)
- Attacker page: a separate Windows VM on `192.168.205.128` serving the PoC page on port `13337`
- Rebinding hostname: `http://7f000001.c0a8cd80.rbndr.us:13337/` (`127.0.0.1` + `192.168.205.128`)
- Browser: Chrome on the Windows host

Observed result in DevTools:
- request URL: `http://7f000001.c0a8cd80.rbndr.us:13337/mcp?cb=...`
- request method: `POST`
- status: `200 OK`
- remote address: `127.0.0.1:13337`
- request headers included:
  - `Host: 7f000001.c0a8cd80.rbndr.us:13337`
  - `Origin: http://7f000001.c0a8cd80.rbndr.us:13337`
  - `Referer: http://7f000001.c0a8cd80.rbndr.us:13337/`
- response body was the full `tools/list` JSON from IDA MCP (`63668` bytes)

So this is not just a cross-origin fetch case. The attacker page becomes same-origin with the rebinding hostname and can read responses from the localhost listener unless the server actively validates `Host` / browser origin context.

I intentionally used the benign `tools/list` method for the repro. With the patch applied, this same request path should be rejected before any MCP method executes.

ZX41R:
Closing the loop with concrete artifacts.

I reproduced this in a real browser on Windows with IDA MCP bound to `127.0.0.1:13337` on the host and the attacker page served from a separate Windows VM on `192.168.205.128:13337` via `7f000001.c0a8cd80.rbndr.us:13337`.

The rebinding page was able to issue a same-origin `POST /mcp` that resolved to `127.0.0.1:13337` while carrying the rebinding `Host` / `Origin`, and I also verified a benign state-changing call with `append_comments` at `0x3420`; the comment rendered in the pseudocode view.

Artifacts:
- DevTools capture: https://github.com/ZX41R/ida-pro-mcp/blob/artifacts/pr-352-evidence/artifacts/pr-352/evidence-01_devtools-request-rebound-to-localhost.png
- IDA pseudocode capture: https://github.com/ZX41R/ida-pro-mcp/blob/artifacts/pr-352-evidence/artifacts/pr-352/evidence-02_ida-pseudocode-comment-poc-confirmed.png
- PoC HTML used for the browser-side request loop: https://github.com/ZX41R/ida-pro-mcp/blob/artifacts/pr-352-evidence/artifacts/pr-352/evidence-03_poc_index_append_comments_poc_confirmed.html

That is the boundary this patch is addressing: browser-readable same-origin access to the localhost MCP surface after rebinding, including mutating methods.

mrexodia:
Thanks for your diligence, I hope you understand my skepticism on issues like this. LLMs love to exaggerate security issues as you probably know 😅

PR #357, opened 2026-04-12
Fix MCP output schema wrapping for union object results

packmad:
Fix MCP tool schema generation for return types expressed as unions of object-shaped results.

It avoids wrapping `anyOf` object unions in a top-level result field when runtime responses are already returned as plain dicts and resolves structured output validation failures for tools like `export_funcs` when using formats such as `prototypes`

### DIFF
--- a/src/ida_pro_mcp/ida_mcp/api_core.py
+++ b/src/ida_pro_mcp/ida_mcp/api_core.py
@@ -508,6 +508,7 @@ def int_convert(
                     ascii=ascii_str,
                     binary=bin(value),
                 ),
+                "error": None,
             }
         )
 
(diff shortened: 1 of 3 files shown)


mrexodia:
Do you have a test for this?

packmad:
@mrexodia I've added a regression test.

SemiOT: 
I don’t understand why I’m the only one having these problems (see my previous PRs). Other people should have them too. Maybe it’s because the other agents are less restrictive about the output, whereas I’m using [openai-agents](https://github.com/openai/openai-agents-python), which are very strict. 
BTW, you can add it in the README if you want, now this library works very well with your project.

mrexodia:
I will add it myself later, thanks for adding tests.

packmad:
@mrexodia I also added an unrelated fix, sorry did not open another PR, but it's super quick to review.

 I added `"error": None` to the success case of `int_convert`

mrexodia:
Thanks!

PR #363, opened 2026-04-14
Use environment variables for MCP host and port

Byzero512:
Updated host and port initialization to use environment variables. use IDAMCP_HOST and IDAMCP_PORT to set host/port for server.

### DIFF
--- a/src/ida_pro_mcp/ida_mcp.py
+++ b/src/ida_pro_mcp/ida_mcp.py
@@ -4,6 +4,7 @@ This file serves as the entry point for IDA Pro's plugin system.
 It loads the actual implementation from the ida_mcp package.
 """
 
+import os
 import sys
 import idaapi
 import ida_kernwin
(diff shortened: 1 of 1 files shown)


mrexodia:
This will not work how you think, If the user changes the settings it will use the setting instead of the environment variable. What is the use case for this anyway?

Byzero512:
start multiple ida sessions with gui in one click

Byzero512:
and if user change the settings, it should use user's settings.

mrexodia:
The commonly-accepted priority order is:
1. command line arguments
2. environment variables
3. user settings
4. defaults

I can take the PR if it is in that order, otherwise I will close it. Additionally I think you are most likely using the wrong tool for the job, probably you want to use the headless idalib MCP instead, that way you can control everything exactly (and have multiple databases open)

Byzero512:
I simply set the priority of `environment variables` higher than that of `defaults`. I think it basically meet your requirements.

mrexodia:
You are banned.

PR #364, opened 2026-04-14
Add --profile flag to idalib-mcp

### LINKED ISSUE #49: Introduce `--profile` argument to allow selecting available functions
it can be difficult to judge which tools are helpful and which tools confuse the LLMs. Additionally there is #46 which introduces potentially dangerous functions. I think adding `--profile myprofile.txt` where `myprofile.txt` contains a list of whitelisted functions could help with this. Some functions can be marked as `@unsafe` and be invisible per default.

NeKroFR:
Closes #49.

`config.html` covers tool selection already, but only inside the plugin.
It stores the selection in a per-IDB netnode and needs the GUI to edit.
`idalib-mcp` has neither, and a netnode isn't something you can commit to
a repo or send to a teammate.

This PR adds a whitelist, same format on both sides.

- `idalib-mcp --profile my_profile.txt` filters the registry at startup.
  Management tools (`idalib_open`, etc...) are kept so you can't lock
  yourself out and unknown names log a warning.
- `config.html` gets an Export link and an Import Profile textarea, so
  you can pull your current selection out as a file, or paste one in.
- I added two examples presets: `profiles/readonly.txt` (35 non-mutating tools)
  and `profiles/triage.txt` (9 tools for a first-pass look).

Why bother when the UI already works? Every registered tool ships its
schema into the model prompt every turn. A 9-tool surface uses roughly
10x fewer tokens than the full set, and small models stop picking the
fancy tool when it isn't on the list. A file is also something you can
commit and share; a netnode isn't.

I also added 12 new tests in `test_profile.py`. And ran the full testsuite on IDA 9.3 still passes.

### DIFF
--- a/profiles/readonly.txt
+++ b/profiles/readonly.txt
@@ -1,0 +1,53 @@
+# Read-only profile for ida-pro-mcp
+#
+# Exposes analysis tools that never mutate the IDB: decompilation, disassembly,
+# xrefs, queries, reads. No renaming, commenting, typing, or patching.
+#
+# Use with: idalib-mcp --profile profiles/readonly.txt <binary>
+
+# --- Core metadata & listings ---
+server_health
+server_warmup
+list_funcs
+list_globals
+imports
+imports_query
+lookup_funcs
+func_query
+entity_query
+int_convert
+find_regex
+
+# --- Analysis (decomp / disasm / xrefs) ---
+decompile
+disasm
+func_profile
+analyze_batch
+xrefs_to
+xref_query
+xrefs_to_field
+callees
+basic_blocks
+find_bytes
+find
+insn_query
+export_funcs
+callgraph
+
+# --- Composite analyses ---
+analyze_function
+analyze_component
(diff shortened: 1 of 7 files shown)


mrexodia:
Awesome, thanks! Which local models did you have success with and what kind of things are you trying to accomplish with the MCP server?

NeKroFR:
I used it only with Claude Opus and Haiku to ensure the feature was working properly. I didn't use it in a big workflow with a lot of agents yet. The goal is mainly to cut token usage and to work better with small context models by restricting `tools/list` so small models don't get overwhelmed.


If we want to test it on a small model, it could be interesting to see how Qwen3 or Llama 3.2 handles reversing with and without a profile. But I think we would need a harder binary than `crackme03.elf` to benchmark properly. If you want I can also add a harder crackme in `tests/`, it might could be useful for benchmarks.

mrexodia:
Yeah would be great if you have some more difficult crackmes for testing!

PR #366, opened 2026-04-15
Fix schema-invalid truncation in large tool outputs

packmad:
This fixes a failure where large structuredContent responses (e.g. `disasm`) were preview-truncated by appending a `{"_truncated": "..."}` sentinel into lists. For tools with strict JSON schemas (`additionalProperties: false` on list item objects), that sentinel makes the response invalid and the client rejects it.

The output limiter now truncates lists by slicing only and relies on the existing `_meta.ida_mcp` truncation metadata + `download_hint text` to indicate that full output is available via download, without breaking schema validation.

### DIFF
--- a/src/ida_pro_mcp/ida_mcp/rpc.py
+++ b/src/ida_pro_mcp/ida_mcp/rpc.py
@@ -48,17 +48,14 @@ def _truncate_value(value: Any, depth: int = 0) -> Any:
         return value[:OUTPUT_LIMIT_PREVIEW_STR_LEN] + f"... [{len(value)} chars total]"
 
     if isinstance(value, list):
-        truncated_list = [
+        # IMPORTANT: Do not inject sentinel objects like {"_truncated": "..."} into lists.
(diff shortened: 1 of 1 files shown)


mrexodia:
I had a suspicion that this might break, but it also returned a list previously so 🤷🏻‍♂️

mrexodia:
Honestly I'm considering to drop structured outputs completely, they don't serve much of a purpose anyway outside of making everything more painful than it has to be...

packmad:
I submit PRsafter thoroughly testing them on at least a hundred malware samples, where I was indeed encountering those errors before. So don't worry, they should all be fairly well tested!

Dropping structured outputs is probably not a good idea: recent papers show they act as a constraint mechanism that reduces the model's search space during generation, which reduces hallucinations.

mrexodia:
Awesome! Would you be open to share some more information about your testing setup/prompts/samples?

packmad:
OFC. I sent you an email.

Issue #34, opened 2025-04-05
Add proper handling for disassembly/decompilation tokens


mrexodia:
IDA has an undocumented format of tokens for their disassembly/decompilation. This might be helpful when generating pseudocode/disassembly because we can add comments with different number bases and translate labels to work better for analysis.

williballenthin:
maybe related code here: https://gist.github.com/williballenthin/466eb28679d30e212ffac57e4a9ceaa5

NeKroFR:
In #367, I added labels, resolved refs ({addr, name}), and anterior/regular/repeatable/posterior comments on every disasm line, idk if it's enough to close this, or if we still need the synthesized multi-base operand annotations (hex/dec/char), I can do it if you want it will be a small PR.

PR #367, opened 2026-04-15
Enrich disasm/decompile output with labels, comments, and refs

### LINKED ISSUE #35: Improve disassembly
Currently the disassembly does not have support for labels and sometimes the LLM fails to resolve function/global destinations.

NeKroFR:
Closes #35. I can take a look at #34 in a follow-up if this is relevant.

`disasm()` was returning per-line `{addr, instruction}` and `decompile()` just rendered text. `DisassemblyLine` already had `label` and `comments` as `NotRequired`, nothing populated them. So the model sees `jz short loc_127F` and has to round-trip through `xrefs_to`/`get_name`.

Each disasm line now gets:
- `label` from `ida_name.get_ea_name(ea)`.
- `comments` from regular + repeatable inline and multi-line anterior/posterior extras, ordered anterior => inline => posterior.
- `refs` walking `CodeRefsFrom`/`DataRefsFrom`, each as `{addr, name}`.
Gotcha: `ida_funcs.get_func_name` returns the *containing* function, so a naive impl resolves `jz short loc_127F` inside `main` as `main`.

`DecompileResult` gets a function-level `refs` list from a `ctree_visitor_t` walking `cot_obj` nodes, deduped by ea. The visitor runs inside an isolated `try` so a Hex-Rays hiccup can't wipe out good pseudocode.

When a ref's target is `is_strlit`, decoded contents are attached as `string`. The codec comes from the low two bits of `strtype` so UTF-16/32 literals don't come back as mojibake.

All new fields are `NotRequired` and only emitted when non-empty. The `instruction` text and the `code` field are untouched, so existing substring assertions keep passing.

I didn't do per-pseudocode-line ref mapping walking `strvec_t` yet (function-level is enough for the model, per-line is bigger, but I can do that if you want). I also skipped the multi-base immediate comments from #34 because `generate_disasm_line` already respects operand display.

I also added 9 new tests in `test_api_analysis.py` on `crackme03.elf` to cover labels, call-target resolution, the containing-function-leak regression guard, data refs, string decoding, regular vs repeatable comments, multi-line anterior extras with ordering pinned, and two decompile-side assertions.

### DIFF
--- a/src/ida_pro_mcp/ida_mcp/api_analysis.py
+++ b/src/ida_pro_mcp/ida_mcp/api_analysis.py
@@ -36,6 +36,7 @@ from .utils import (
     extract_function_constants,
     Argument,
     DisassemblyFunction,
+    Ref,
     Xref,
     BasicBlock,
     StructFieldQuery,
@@ -50,6 +51,7 @@ from . import compat
 class DecompileResult(TypedDict):
     addr: str
     code: str | None
+    refs: NotRequired[list[Ref]]
     error: NotRequired[str]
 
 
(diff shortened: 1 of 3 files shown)


mrexodia:
🥳

PR #372, opened 2026-04-16
feat(api_core): add search_text tool

### LINKED ISSUE #67: [Feature Request] Search text function
I often find AI Agents using regex in the ```list_strings``` output which seems unsupported/broken? I have to manually search a lot and copy/paste the results. I also find the "Search Text" function in IDA much more useful, it often contains things not in the "Strings" window. Is it possible to add this as an MCP tool?

NeKroFR:
Closes #67.

Adds `search_text`, an MCP tool that does what IDA's native "Search Text" dialog does: regex-match against disassembly text and comments, with pagination.

## API

```python
search_text(
    pattern,                     # case-insensitive regex
    limit=30,                    # clamped to [1, 500]
    offset=0,
    include="all",               # "disasm" | "comments" | "all"
    code_only=True,              # executable segments only
) -> {n, matches, cursor, error?}
```

Each match: `{addr, text, kind: "disasm" | "comment", function?}`. An instruction matched in both disasm and a comment becomes two records (pagination key is `(addr, kind)`). Same `cursor` shape as `find_regex`. Invalid regex or invalid `include` surface as `{error}` instead of raising.

## Implementation

Iterates `idautils.Segments()` => `idautils.Heads()`, pulls `ida_lines.generate_disasm_line` + `tag_remove` for the disasm side, and gathers anterior/regular/repeatable/posterior comments via a small `_iter_head_comments` helper. No new deps, ~100 LoC in `api_core.py`.

## Tests

8 new tests in `test_api_core.py`: no-match envelope, invalid regex, invalid include, match shape, `\bcall\b` finds call sites, pagination cursor advances without duplicate `(addr, kind)`, `limit` respected, `include="disasm"` excludes comments.

### DIFF
--- a/src/ida_pro_mcp/ida_mcp/api_core.py
+++ b/src/ida_pro_mcp/ida_mcp/api_core.py
@@ -10,6 +10,7 @@ import idaapi
 import ida_funcs
 import ida_hexrays
 import ida_lines
+import ida_search
 import ida_segment
 import idautils
 import ida_loader
@@ -114,16 +115,21 @@ class FindRegexResult(TypedDict, total=False):
     error: str | None
 
 
-class SearchTextMatch(TypedDict, total=False):
-    addr: str
-    text: str
+class SearchTextLine(TypedDict, total=False):
     kind: str  # "disasm" | "comment"
+    text: str
+
+
+class SearchTextHit(TypedDict, total=False):
+    addr: str
     function: str
+    segment: str
+    matches: list[SearchTextLine]
 
 
 class SearchTextResult(TypedDict, total=False):
     n: int
-    matches: list[SearchTextMatch]
+    hits: list[SearchTextHit]
     cursor: dict[str, Any]
     error: str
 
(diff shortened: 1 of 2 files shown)


mrexodia:
This seems very inefficient, needs more research

mrexodia:
I think adding the traces to the IDB should be done automatically and as a 'first-class citizen' (likely in some compressed form). Then the JSONL should be exportable from the actual source of truth. This way if you share the IDB you also automatically share the traces. Will do some research on this

mrexodia:
Verdict

 The PR’s current approach is functional, but not the most efficient:

 - it walks every head
 - does multiple Python→IDA calls per EA
 - regex-matches in Python
 - extracts comments eagerly even when there are none

 That’s expensive.

 Best practical design

 I’d implement it as a hybrid:

 1. Use IDA’s native listing search to find candidate addresses fast:
     - ida_search.find_text(...)
     - this runs the scan in C++, not in Python
 2. Only for matched addresses, render the listing once:
     - ida_lines.generate_disassembly(...)
     - then extract:
           - matching snippet(s)
           - whether match was in listing vs comment
           - useful agent context

 That gives you most of the speedup without losing usability.

 ---

 Why ida_search.find_text() is the right primitive

 IDA already has a built-in text search engine:

 - find_text(start_ea, y, x, ustr, sflag) in search.hpp
 - implementation in base/search.cpp

 It supports:
 - case-sensitive / insensitive
 - regex via SEARCH_REGEX
 - quiet mode via SEARCH_NOSHOW

 And importantly, it searches the rendered listing text in C++ using linearray_t, instead of doing:
 - generate_disasm_line() per head
 - get_cmt() per head
 - get_extra_cmt() loops per head

 So it is fundamentally a better engine for this job.

 ---

 One caveat

 ida_search.find_text() returns only:
 - the EA

 It does not directly return:
 - which line matched
 - whether it was a comment or disasm
 - the snippet text

 So I would not use it alone if you want agent-friendly output.

 That’s why the hybrid approach is better.

 ---

 What I would change in the API

 Since schema can change, I’d optimize for agent usefulness, not line-perfect exhaustiveness.

 Instead of returning a flat per-line match list, return one hit per EA with compact context.

 Example:

 ```json
   {
     "hits": [
       {
         "addr": "0x401234",
         "function": "decrypt",
         "segment": ".text",
         "matches": [
           { "kind": "disasm", "text": "call sub_401000" },
           { "kind": "comment", "text": "decrypt key buffer" }
         ]
       }
     ],
     "next_cursor": "0x401234",
     "done": false
   }
 ```

 For agents, this is usually better:
 - fewer duplicates
 - easier to reason about
 - enough context to decide next action
 - easy to follow up with disasm/decompile tools

 ---

 Recommended implementation strategy

 ### 1. Add a literal-vs-regex mode

 Right now PR 372 is regex-only.

 That’s not ideal for performance.

 IDA’s search engine has separate fast paths for:
 - literal substring search
 - regex search

 So I’d make it:

 - regex: bool = False
 - default to literal substring search
 - only enable SEARCH_REGEX when requested

 That alone should help a lot for typical agent usage.

 ---

 ### 2. Search with ida_search.find_text()

 Pseudo-flow:

 ```python
   sflag = ida_search.SEARCH_DOWN | ida_search.SEARCH_NOSHOW
   if case_sensitive:
       sflag |= ida_search.SEARCH_CASE
   if regex:
       sflag |= ida_search.SEARCH_REGEX
 ```

 Then loop with find_text(...) from the current cursor.

 ---

 ### 3. Restrict to executable segments yourself

 find_text() doesn’t give you an end bound in the Python wrapper, so for code_only=True I’d iterate exec segments manually:

 - start search from current segment start or cursor
 - call find_text(...)
 - accept only if result is within that segment
 - otherwise move to next exec segment

 That still avoids the full Python-side head scan.

 ---

 ### 4. For each hit EA, call generate_disassembly() once

 Use:

 - ida_lines.generate_disassembly(ea, max_lines, as_stack=False, notag=False)

 This gives you the rendered tagged lines for the item.

 Then:
 - strip tags for matching/snippets
 - inspect tags to classify comments

 IDA’s own example jump_next_comment.py shows how to detect comment lines using:
 - SCOLOR_REGCMT
 - SCOLOR_RPTCMT
 - SCOLOR_AUTOCMT

 So you can classify lines as:
 - comment
 - listing or disasm

 without doing separate get_cmt() / get_extra_cmt() calls.

 ---

 ### 5. Return one result per EA

 For each candidate EA:
 - collect the matching rendered lines
 - filter by include=all|comments|disasm
 - if any remain, emit a single hit object

 That avoids hard pagination problems around multiple matches in one item.

 ---

 Why this is better than the current PR

 ### Current PR

 - O(all heads)
 - many Python/IDA crossings
 - comment APIs called on most EAs
 - regex always in Python

 ### Proposed hybrid

 - native C++ search over rendered listing
 - one render per actual hit
 - simpler, agent-oriented output
 - literal search can be fast by default
 - classification still possible

 ---

 If you wanted the absolute simplest/faster version

 If you’re okay with even looser output, the fastest useful version is probably:

 - use only ida_search.find_text()
 - return:
     - addr
     - function
     - maybe one best-effort preview line from generate_disassembly()

 No exhaustive per-line enumeration at all.

 That would be very agent-usable already.

---

 I would not keep the current implementation as-is

 At minimum, even if you stayed close to PR 372, I’d still say:
 - don’t scan every head
 - don’t fetch comments for every EA
 - don’t force regex for all searches

 So yes: there is a more efficient way to implement PR 372, and I think the best version is:

 │ ida_search.find_text() for candidate discovery + generate_disassembly() for hit context.

 If you want, I can sketch the exact replacement function signature and implementation shape for api_core.py.

NeKroFR:
I've pushed a rewrite that follows your hybrid design:

- `ida_search.find_text()` for candidate discovery (C++), with `SEARCH_DOWN | SEARCH_NOSHOW + SEARCH_CASE/SEARCH_REGEX` as flags.
- One `ida_lines.generate_disassembly()` render per hit; comment vs disasm classified via `SCOLOR_REGCMT/RPTCMT/AUTOCMT` tags no more per-EA `get_cmt/get_extra_cmt`.
- regex: `bool = False` literal substring by default.
- code_only enforced by manual exec-segment iteration with per-segment bound checks.
- Output reshaped to one hit per EA: `{addr, function, segment, matches:[{kind,text}]}`, cursor is an address.

No more full-head walk, no eager comment fetching, no forced regex. Let me know if the shape or any detail needs another pass.

mrexodia:
Thanks!

PR #374, opened 2026-04-16
feat(trace): store tools/call trace in IDB + idalib dump script

### LINKED ISSUE #189: Add tracing functionality
It would be helpful to save a trace of all the called tools input/output pairs in the IDB/to a file. This might help shed some light on buggy tool calls for clients that do not support detailed traces (pretty much non of them do).

NeKroFR:
Closes #189.

Appends one JSON line per `tools/call` to a file.

Each record: `{ts, tool, arguments, structuredContent, isError, duration_ms, truncated?, full_result_size?}`. `@unsafe` tool arguments (`py_eval`, `py_exec_file`) are redacted by default. Pass `--trace-verbose` to keep them as-is.

## Enable

```sh
# CLI
idalib-mcp --trace-file /tmp/trace.jsonl [--trace-verbose]

# Plugin (no CLI)
IDA_MCP_TRACE_FILE=/tmp/trace.jsonl ida ...
```

## Read

```sh
cat /tmp/trace.jsonl | jq .
```

The issue mentions "IDB/to a file". I went with the file path since it's simpler and covers the stated use case. An IDB netnode backend would be a natural follow-up so the trace travels with the database, maybe with `trace_list` / `trace_clear` tools on top. I can do that if you prefer.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -40,6 +40,7 @@ build-backend = "setuptools.build_meta"
 ida-pro-mcp = "ida_pro_mcp.server:main"
 idalib-mcp = "ida_pro_mcp.idalib_server:main"
 ida-mcp-test = "ida_pro_mcp.test:main"
+ida-mcp-trace-dump = "ida_pro_mcp.trace_dump:main"
 
 [tool.pyright]
 include = ["src/ida_pro_mcp"]
--- a/src/ida_pro_mcp/ida_mcp/__init__.py
+++ b/src/ida_pro_mcp/ida_mcp/__init__.py
@@ -37,7 +37,7 @@ from . import api_resources
 from . import api_survey
 from . import api_composite
 from . import api_discovery
-from . import trace as trace  # registers the trace_clear tool; must precede http import
+from . import trace as trace
 
 # Re-export key components for external use
 from .sync import idasync, IDAError, IDASyncError, CancelledError
@@ -46,9 +46,8 @@ from .http import IdaMcpHttpRequestHandler
 from .api_core import init_caches
 from .api_discovery import set_local_instance
 
-# Activate tracing if IDA_MCP_TRACE_FILE is set (env-var opt-in for the
-# in-IDA plugin, which has no CLI).
-trace.configure_from_env()
+# Tracing is always on: every tools/call is recorded into the IDB netnode.
+trace.configure_idb()
 
 __all__ = [
     # Infrastructure modules
--- a/src/ida_pro_mcp/ida_mcp/__init__.py
+++ b/src/ida_pro_mcp/ida_mcp/__init__.py
@@ -37,6 +37,7 @@ from . import api_resources
 from . import api_survey
 from . import api_composite
 from . import api_discovery
+from . import trace as trace  # registers the trace_clear tool; must precede http import
 
 # Re-export key components for external use
 from .sync import idasync, IDAError, IDASyncError, CancelledError
@@ -44,7 +45,6 @@ from .rpc import MCP_SERVER, MCP_UNSAFE, tool, unsafe, resource
 from .http import IdaMcpHttpRequestHandler
 from .api_core import init_caches
 from .api_discovery import set_local_instance
-from . import trace as trace
 
 # Activate tracing if IDA_MCP_TRACE_FILE is set (env-var opt-in for the
 # in-IDA plugin, which has no CLI).
--- a/src/ida_pro_mcp/ida_mcp/__init__.py
+++ b/src/ida_pro_mcp/ida_mcp/__init__.py
@@ -44,6 +44,11 @@ from .rpc import MCP_SERVER, MCP_UNSAFE, tool, unsafe, resource
 from .http import IdaMcpHttpRequestHandler
 from .api_core import init_caches
 from .api_discovery import set_local_instance
+from . import trace as trace
+
+# Activate tracing if IDA_MCP_TRACE_FILE is set (env-var opt-in for the
+# in-IDA plugin, which has no CLI).
+trace.configure_from_env()
 
 __all__ = [
     # Infrastructure modules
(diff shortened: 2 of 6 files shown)


mrexodia:
Best approach

 Use a dedicated netnode-backed, append-only log made of compressed segments, not one ever-rewritten blob.

 ### Why

 From IDA’s netnode implementation:

 - MAXSPECSIZE = 1024 bytes per supval (include/netnode.hpp)
 - setblob() is implemented as multiple contiguous supvals
 - getblob() reads until the first missing index
 - setblob() also deletes trailing chunks after the end of the blob it writes (netnode/nn19.cpp)

 So if you store “the whole trace” as one blob at index 0, every append becomes:

 1. read all
 2. decompress all
 3. append one record
 4. recompress all
 5. rewrite all

 That’s O(n) per append and gets worse over time.

 Recommended layout

 Use one node, e.g.:

 - netnode name: "$ ida_mcp.trace"

 Use 3 tags:

 - meta tag: store version / next segment start / next segment id
 - index tag: map segment_id -> start_index
 - data tag: store compressed blob for each segment

 ### Important trick

 Store each compressed segment as its own blob and leave one empty index gap after it.

 Because getblob(start, tag) stops at the first missing index, that gap cleanly delimits segments.

 That gives you true append-only behavior with no rewrites.

 Compression format

 Use gzip per segment.

 Why gzip:

 - easy from Python stdlib
 - self-contained members
 - restart-safe
 - corruption/truncation only affects the last segment
 - easy export later

 I would not keep one long-lived streaming compressor state across sessions. That’s awkward to resume and fragile.

 Concrete design

 ### Metadata

 For example under altvals:

 - meta[0] = version
 - meta[1] = next_chunk_start
 - meta[2] = next_segment_id
 - optionally meta[3] = total_records

 ### Append

 For each flush:

 1. build JSONL bytes for 1..N trace records
 2. gzip.compress(...)
 3. write with node.setblob(member, start, TAG_DATA)
 4. store segment_id -> start in index tag
 5. advance next_chunk_start by:
     - ceil(len(member) / 1024) + 1
     - the +1 is the delimiter gap

 ### Read/export

 Iterate segment ids in order:

 - start = node.altval(seg_id, TAG_INDEX)
 - member = node.getblob(start, TAG_DATA)
 - gzip.decompress(member)

 ### Clear

 - easiest: node.kill()

 Why this is better than one contiguous compressed stream

 A single concatenated stream can work, but segmented blobs are better because:

 - no full rewrite ever
 - simpler recovery
 - simpler iteration
 - easier trace_list, trace_export, trace_clear
 - easier future size caps / pruning
 - avoids “read entire blob to find tail” behavior

 If you really want a single logical stream, you can still think of it as a stream of gzip members, but I’d store them as separate netnode blobs with an index,
 not as one giant contiguous blob.

 One more important point: main thread

 If this is stored in the IDB, all netnode operations must run on the IDA main thread.

 So in ida-pro-mcp, the actual append/read/clear helpers should be wrapped through @idasync, even if the tracing middleware itself runs outside that thread.

 That’s probably the biggest implementation gotcha.

 Practical recommendation for this repo

 I’d implement:

 - keep current file backend as-is
 - add optional IDB backend, something like:
     - IDA_MCP_TRACE_IDB=1 or --trace-idb
 - batch writes a little:
     - flush every N records or every ~32–64 KB
     - plus flush on shutdown

 Batching matters because netnode writes are journaled/undo-tracked and per-call writes will bloat the IDB faster.

 Minimal sketch

 Something like:

 ```python
   NODE = "$ ida_mcp.trace"
   TAG_META = "M"
   TAG_INDEX = "I"
   TAG_DATA = "D"
   CHUNK = 1024
 ```

 Append logic:

 ```python
   @idasync
   def append_trace_segment(payload: bytes):
       node = ida_netnode.netnode(NODE, 0, True)

       version = node.altval(0, TAG_META) or 1
       start = node.altval(1, TAG_META) or 0
       seg_id = node.altval(2, TAG_META) or 0

       member = gzip.compress(payload, mtime=0)
       node.setblob(member, start, TAG_DATA)
       node.altset(seg_id, start, TAG_INDEX)

       used_chunks = (len(member) + CHUNK - 1) // CHUNK
       node.altset(0, version, TAG_META)
       node.altset(1, start + used_chunks + 1, TAG_META)  # +1 gap
       node.altset(2, seg_id + 1, TAG_META)
 ```

 Bottom line

 Best implementation:
 Use a dedicated netnode with append-only compressed segments, indexed by altvals, with one empty chunk gap between segments.

 That fits IDA’s blob semantics much better than a single rewritten blob, and it gives you a clean path to trace_list, trace_export, and trace_clear.

 If you want, I can prototype this backend in ida_pro_mcp next.

NeKroFR:
Implemented as you described. IDB backend behind `--trace-idb` / `IDA_MCP_TRACE_IDB=1`, kept previous backend.

Netnode `$ ida_mcp.trace`, three tags (META/INDEX/DATA), gzipped batch per segment, +1 supval gap. Batching at 256 records or 64 KB, flush on shutdown. Added `trace_clear` tool and `trace_list/trace_export` can follow. All netnode ops via `@idasync`.

One detail: segment write + all meta bumps happen in a single `@idasync` job, and `clear()` holds a flush lock, so a concurrent clear can't land between partial updates.

mrexodia:
Just FYI I still need some time to review this properly, but some notes:
- The trace should always be stored in the IDB (no option to disable)
- Nothing should be 'redacted'
- There shouldn't be a trace_clear tool
- Instead of (or in addition to, not 100% sure) I think we should switch the IDB to be the source of truth and have a script that uses idalib to dump a JSONL from an IDB

NeKroFR:
Latest commit addresses all four points:
 - Tracing is now always-on in the IDB
 - Arguments are stored verbatim (no redaction)
 - trace_clear tool removed
 - Added ida-mcp-trace-dump (uses idalib) to export an IDB's trace as JSONL, the IDB is now the source of truth

Take your time on the review, it is better to land the right design than to rush it. Happy to iterate on anything.

mrexodia:
I don't want to keep it open forever, will test it together with the other changes and open an issue with stuff I find. Thanks a lot!

NeKroFR:
Don't hesitate to ping me on the issues, happy to help.

PR #378, opened 2026-04-21
fix: make caller graph detection architecture agnostic

pyuysig:
## Summary
- replace architecture-specific `NN_call*` checks with `idaapi.is_call_insn(insn)`
- use `CodeRefsFrom` to collect call destinations for caller/callee graph queries
- add regression coverage for the architecture-agnostic call detection path

## Why
The old caller/callee implementation was tied to x86-style `NN_call*` instruction IDs. On non-x86 targets such as MIPS, live headless runs could still observe real code xrefs, but `get_callers`/`get_callees` returned empty results because the callsite filter never matched. This broke downstream graph consumers even when the underlying IDA database had the route information.

## Validation
- local unit tests in this repo
- live downstream validation through a rebuilt `sdk-idapro` managed bundle on a MIPS target, where `get_callers(0x10001370)` began returning `main` and downstream route witnesses became available

### DIFF
--- a/src/ida_pro_mcp/ida_mcp/utils.py
+++ b/src/ida_pro_mcp/ida_mcp/utils.py
@@ -1202,6 +1202,42 @@ def get_all_comments(ea: int) -> dict:
     return comments
 
 
+def _decode_insn_at(ea: int):
+    """Decode the instruction at ea, returning None when decode fails."""
+    insn = idaapi.insn_t()
+    if idaapi.decode_insn(insn, ea) == 0:
+        return None
+    return insn
+
+
+def _is_call_instruction(insn: idaapi.insn_t) -> bool:
+    """Architecture-agnostic call instruction detection with legacy fallback."""
+    try:
+        return bool(idaapi.is_call_insn(insn))
+    except Exception:
+        return insn.itype in [idaapi.NN_call, idaapi.NN_callfi, idaapi.NN_callni]
+
+
+def _direct_call_targets(ea: int, insn: idaapi.insn_t) -> list[int]:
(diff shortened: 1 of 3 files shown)


src/ida_pro_mcp/ida_mcp/utils.py
```diff
@@ -1202,6 +1202,42 @@ def get_all_comments(ea: int) -> dict:
...
 
+def _decode_insn_at(ea: int):
+    """Decode the instruction at ea, returning None when decode fails."""
+    insn = idaapi.insn_t()
+    if idaapi.decode_insn(insn, ea) == 0:
+        return None
+    return insn
+
+
+def _is_call_instruction(insn: idaapi.insn_t) -> bool:
```
mrexodia, line 1213: Get rid of this and use `idaapi.is_call_insn` directly

src/ida_pro_mcp/ida_mcp/utils.py
```diff
@@ -1202,6 +1202,42 @@ def get_all_comments(ea: int) -> dict:
...
+    """Architecture-agnostic call instruction detection with legacy fallback."""
+    try:
+        return bool(idaapi.is_call_insn(insn))
+    except Exception:
+        return insn.itype in [idaapi.NN_call, idaapi.NN_callfi, idaapi.NN_callni]
+
+
+def _direct_call_targets(ea: int, insn: idaapi.insn_t) -> list[int]:
```
mrexodia, line 1221: Same here, just inline it at the call site there is only one user

mrexodia (changes requested):

PR #383, opened 2026-04-23
Respect env IDA_MCP_URL when setting download base url

Evian-Zhang:
The environment `IDA_MCP_URL` is used to customize `download_url` string returned by MCP (code resides in [here](https://github.com/mrexodia/ida-pro-mcp/blob/89a0016765ee62dfd9414c2db4c9d9d6a7a62319/src/ida_pro_mcp/ida_mcp/rpc.py#L17)). The PR #375 update download url by the host and port passed in cmdline regardless of the environment. This PR makes the behavior respect the environment.

The motivation for this is that, the MCP server may run behind complex networks such as NAT. As a result, the server may listens to `192.168.1.2:2333`, while the client at the other network may access it by `10.8.20.2:3444`. The `download_url` is used to indicate the client about how to access the truncated data, while the client cannot access `192.168.1.2:2333` at all. So this PR make it possible that we can customize the literal string "download_url" returned by MCP server (this does NOT change the actual socket that server listens to).

### DIFF
--- a/src/ida_pro_mcp/ida_mcp/api_discovery.py
+++ b/src/ida_pro_mcp/ida_mcp/api_discovery.py
@@ -12,9 +12,11 @@ import subprocess
 import sys
 import threading
 import time
+from collections import OrderedDict
 from typing import Annotated, NotRequired, TypedDict
 
 from .rpc import tool, MCP_SERVER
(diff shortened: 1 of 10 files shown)


mrexodia:
@codex

PR #395, opened 2026-05-03
Support disabling expensive tools

momo5502:
Certain tools, especially those that search/find text, can take a long time to complete. Not exposing them to agents can make sense, depending on the binary that's being analyzed.

Using `http://127.0.0.1:13337/mcp?disable=expensive` or `http://127.0.0.1:13337/mcp?ext=dbg&disable=expensive`, the tools `find`, `find_regex` and `search_text` can now be disabled.


momo5502:
config.html fits my needs.

PR #396, opened 2026-05-03
Fix open_file ida executable path on linux

D00Movenok:
Found that in IDA Python `sys.executable` returns the Python executable on Linux, while on other OSes it works correctly. This behaviour breaks the `open_file` tool. In this PR, it now reads the IDA path from `/proc/self/cmdline` on Linux.

On linux:
image

On OSX:
image

On Windows:
image


mrexodia:
Thanks!

PR #407, opened 2026-05-09
Fix call_stack deadlock on reentrant @idasync (closes #406)

### LINKED ISSUE #406: `_sync_wrapper` deadlocks IDA main thread on reentrant `@idasync` (race in `call_stack` LifoQueue)
## Summary

`_sync_wrapper` in `sync.py` uses a process-wide `call_stack = queue.LifoQueue()` with **blocking `get()`** to track which tool is currently running on the IDA main thread. A reentrant `@idasync` call from within the main thread (e.g. `_netnode_flush_segment` from `trace.py` triggered while another tool is mid-execution) drains the queue between the outer tool's `put()` and its `finally: call_stack.get()`. The outer tool then blocks forever on `Queue.get(block=True)` against an empty queue, freezing IDA's main thread. All subsequent `@idasync` HTTP requests pile up in `execute_sync`'s queue and never return.

This is a different mechanism than #217 (which describes a UI-dialog block on the main thread). Here the UI is idle — the deadlock is internal to `_sync_wrapper` after the wrapped function has already returned successfully.

## Repro environment

- ida-pro-mcp **2.0.0** (current `main` is functionally identical for these lines)
- IDA Pro 9.x on Windows 10/11
- Python 3.11
- MCP client: Claude Code (any client that issues calls fast enough to overlap with `trace._netnode_flush_segment` works)
- Multiple IDA instances open with this MCP plugin (load increases probability)

## Reproduction

The race fires when:
1. A `@tool @idasync` function is mid-execution (`runned()` already did `call_stack.put('outer')`).
2. **Inside that function, while still on the main thread**, another `@idasync` function is invoked synchronously. With `idaapi.execute_sync()` called from the main thread, the callback runs **inline** in the same stack — there is no marshalling.
3. The reentrant `runned()` sees `call_stack.empty() == False`, calls `call_stack.get()`, **removes the outer tool's entry**, and raises `IDASyncError("Call stack is not empty while calling _netnode_flush_segment from idb_save")`.
4. The exception is caught in the reentrant `_sync_wrapper`'s `res_container`. Control returns to the outer tool.
5. Outer tool finishes normally. `_sync_wrapper.runned()` reaches its `finally: call_stack.get()`. The queue is now empty. **`get()` with `block=True` (default) waits forever.**
6. IDA's main thread is permanently parked on the `LifoQueue` condition variable. Every subsequent `@idasync` call (`config_json_get`, `idb_meta`, anything) enqueues to `execute_sync` and joins the wait.

I observed this concretely on `idb_save`. The native `ida_loader.save_database()` itself completed in 31 ms with `ok=True` — the freeze was strictly post-save, in the wrapper's `finally`. `trace.py:_netnode_flush_segment` (also `@idasync`) was the reentrant caller because `MCP_SERVER` records every tool invocation into the `$ ida_mcp.trace` netnode.

## Trace evidence

I added per-line instrumentation to `_sync_wrapper` and `idb_save` (file logger with `fsync` per write, plus a worker-side heartbeat). The timeline of the freeze, captured from a real session:

```
1778273149.605817 tid=29348 sync.enter tool='idb_save' keep_batch=False
1778273149.607818 tid=29348 sync.execute_sync.queueing tool='idb_save' on_main=False
1778273149.607818 tid=65520 sync.runned.entered tool='idb_save'
1778273149.608817 tid=65520 sync.batch.before tool='idb_save'
1778273149.608817 tid=65520 sync.batch.after tool='idb_save' old_batch=False
1778273149.609818 tid=65520 sync.ff.call tool='idb_save'
1778273149.609818 tid=65520 idb_save.enter path=''
1778273149.611818 tid=65520 idb_save.save_database.BEFORE
1778273149.615819 tid=65520 sync.callstack.error tool='_netnode_flush_segment' prev='idb_save'   ← reentrant call drains call_stack
1778273149.638818 tid=65520 idb_save.save_database.AFTER ok=True dur_s='0.031'
1778273149.638818 tid=65520 idb_save.return ok=True
1778273149.639817 tid=65520 sync.ff.returned tool='idb_save'
1778273149.639817 tid=65520 sync.batch.restoring tool='idb_save'
                                                 ← main thread hangs here, in call_stack.get()
1778273154.608774 tid=3864 worker.still_waiting tool='idb_save' elapsed='5.0s'
1778273159.611797 tid=3864 worker.still_waiting tool='idb_save' elapsed='10.0s'
...
1778304310.190109 tid=11376 worker.still_waiting tool='config_json_get' elapsed='31034.9s'   ← 8.6 hours later
1778304310.706110 tid=71908 worker.still_waiting tool='config_json_get' elapsed='31035.4s'
```

`sync.runned.exiting` never appears for `tool='idb_save'`. The instrumentation places that record **immediately after** `call_stack.get()` in the `finally` block — so the main thread is provably stuck on that single line. Meanwhile dozens of HTTP-keepalive worker threads enqueue `config_json_get` and accumulate.

## Root cause

`call_stack` is intended as a debug guard: "no two tools should be on the main thread simultaneously." But:

- The check (`empty()` → `get()`) is **not atomic** w.r.t. reentrant entries from the same thread.
- The cleanup `call_stack.get()` in `finally` assumes nobody else has popped the queue.
- Both assumptions break the moment a `@idasync` function calls another `@idasync` function (directly or transitively, e.g. via a hook/save-side-effect).

Either side of the race uses **blocking** `Queue.get()`. The reentrant `get()` happens to find an item (the outer tool's name) and succeeds. The outer `finally`'s `get()` finds nothing and blocks indefinitely.

## Proposed fix

Two-line change: use `get_nowait()` and tolerate `queue.Empty` in both places. The semantics of `call_stack` (debug guard) are preserved: if it's been drained by a reentrant entry, the outer `finally` simply finds nothing to clean up.

```diff
--- a/src/ida_pro_mcp/ida_mcp/sync.py
+++ b/src/ida_pro_mcp/ida_mcp/sync.py
@@
     def runned():
         if not call_stack.empty():
-            last_func_name = call_stack.get()
+            try:
+                last_func_name = call_stack.get_nowait()
+            except queue.Empty:
+                last_func_name = "<empty>"
             error_str = f"Call stack is not empty while calling the function {ff.__name__} from {last_func_name}"
             raise IDASyncError(error_str)
@@
         finally:
             if not (completed and keep_batch):
                 idc.batch(old_batch)
             _sync_state.pre_call_batch = prev_pre_call
-            call_stack.get()
+            try:
+                call_stack.get_nowait()
+            except queue.Empty:
+                pass
```

This eliminates the freeze. It does not address the *root* design issue (a shared mutable LifoQueue is the wrong primitive for this guard — a `threading.local` stack would be more correct, since reentrancy is by definition same-thread), but it removes the failure mode users actually hit.

A more invasive follow-up could:
1. Replace `call_stack` with a `threading.local()` list, or
2. Detect reentrancy explicitly (`if threading.current_thread() is threading.main_thread() and len(call_stack) > 0: just inline-call ff()`) instead of going through `execute_sync` again.

But the two-line patch above is sufficient to keep `idb_save` and other `@idasync` tools from freezing IDA, and it should land cleanly.

## Why this surfaced now

Tracing into the IDB netnode (`trace.configure_idb()` enabled by default in 2.0.0's `__init__.py`) wraps `MCP_SERVER` and writes via `@idasync _netnode_flush_segment`. Any tool whose body indirectly causes a flush before its outer `runned()` exits will trigger this race. `idb_save` is the most reproducible because the save path interacts with hooks; but other tools can hit it when load is high.

## Related

- #217 (open) — main-thread blocking, but on UI dialogs / blocked main thread. Different cause.
- #165 (closed) — `@idaread` nesting deadlock. Closed before `@idasync` unification (#08d08c08).

atimofeev86:
Fixes #406. See that issue for full diagnosis, trace evidence (8.7-hour freeze captured live), and reproduction conditions.

## What this changes

Two-line fix in `_sync_wrapper`: replace blocking `call_stack.get()` with `get_nowait()` + tolerate `queue.Empty` in both the pre-check and the `finally` cleanup. The debug-guard semantics ("no two tools on the main thread simultaneously") are preserved — if a reentrant entry has already drained the queue, the outer `finally` finds nothing to clean up, which is correct.

## Why it's needed

`call_stack` is a `queue.LifoQueue()` shared across the process. The blocking `get()` deadlocks the moment a `@idasync` function transitively invokes another `@idasync` function from the main thread:

1. Outer tool's `runned()` puts its name on `call_stack`.
2. Reentrant call (e.g. `trace._netnode_flush_segment` triggered by tool-call recording into the `$ ida_mcp.trace` netnode) runs **inline** on the main thread because `execute_sync` from the main thread is synchronous.
3. Reentrant pre-check sees `call_stack.empty() == False`, calls blocking `get()`, removes the outer entry, raises `IDASyncError`.
4. Outer tool finishes normally, hits `finally: call_stack.get()` — queue empty — blocks forever.
5. IDA's main thread is parked on the `LifoQueue` condition variable. Every subsequent `@idasync` request piles up in `execute_sync`.

Most reproducible on `idb_save` because the save path interacts with hooks that flush the trace netnode mid-call. Other tools can hit it under load. Default 60s timeout (`sys.setprofile`-based) doesn't apply because the wait is in C, not Python.

## Notes

- Doesn't address the deeper design issue (a shared mutable LifoQueue is the wrong primitive here — `threading.local()` would be more correct, since reentrancy is by definition same-thread). That would be a larger refactor; this PR keeps the diff to the minimum needed to unblock real-world freezes.
- 100% backward compatible: the pre-check still raises `IDASyncError` when the queue is non-empty (now with `<empty>` as the prev-name fallback, which only happens in the race window).

### DIFF
--- a/src/ida_pro_mcp/ida_mcp/sync.py
+++ b/src/ida_pro_mcp/ida_mcp/sync.py
@@ -91,7 +91,13 @@ def _sync_wrapper(ff, keep_batch=False):
 
     def runned():
         if not call_stack.empty():
-            last_func_name = call_stack.get()
+            # Non-blocking: a concurrent reentrant @idasync call from
+            # within another tool's ff() on the same main thread may
+            # have drained the queue between empty() and get().
+            try:
+                last_func_name = call_stack.get_nowait()
+            except queue.Empty:
+                last_func_name = "<empty>"
             error_str = f"Call stack is not empty while calling the function {ff.__name__} from {last_func_name}"
             raise IDASyncError(error_str)
 
@@ -110,7 +116,14 @@ def _sync_wrapper(ff, keep_batch=False):
             if not (completed and keep_batch):
                 idc.batch(old_batch)
             _sync_state.pre_call_batch = prev_pre_call
-            call_stack.get()
+            # Non-blocking: a reentrant @idasync invoked synchronously
+            # inside ff() may have already popped our entry. Default
+            # block=True would freeze the IDA main thread on an empty
+            # queue and hang every subsequent @idasync call.
+            try:
+                call_stack.get_nowait()
+            except queue.Empty:
+                pass
 
     idaapi.execute_sync(runned, idaapi.MFF_WRITE)
     res = res_container.get()


miceiken:
Also having problems with `ida_save` and even `py_eval` of `ida_loader.save_database()`. Thanks for figuring out the bug.

mrexodia:
I think the final fix needs to be different, but this will work for now!

PR #409, opened 2026-05-10
Fix LAN connectivity: unify security checks for non-loopback bindings

### LINKED ISSUE #206: local server to LAN server
It seems that the mcp server after startup can only allow local connections. Is there any other way to allow all machines in the LAN to connect?

Aliang1337:
## Summary

Fixes #206 — when the MCP server is bound to a non-loopback address (e.g. `0.0.0.0` via the IDA config form), the hardcoded localhost-only checks in `_check_host` and `_check_origin` (`http.py`) reject all non-localhost requests with 403.

- Replace `_check_host` and `_check_origin` with `_check_api_request` from zeromcp, which already handles non-loopback bindings correctly (allows any Host header when not bound to loopback, enforces CORS policy for Origin)
- Add LAN-specific test cases for `_check_api_request` with `0.0.0.0` binding

### How it works

When bound to a non-loopback address:
- **MCP API calls** from non-browser clients (no Origin header): work immediately
- **Config page** (`/config.html`): accessible from LAN for viewing
- **Config form submission**: requires CORS policy set to "unrestricted"
- **DNS rebinding protection**: preserved — loopback-bound servers still reject non-loopback Host headers

No changes to `ida_mcp.py` — LAN exposure is expected to go through `idalib-mcp --host 0.0.0.0` or manual config form override.

## Test plan

- [x] All 13 existing + new browser transport guard tests pass
- [x] All 58 related unit tests pass with no regression
- [ ] Manual: configure host to `0.0.0.0` via IDA config form, connect from LAN machine
- [ ] Manual: verify loopback-bound server still rejects external Host headers

### DIFF
--- a/src/ida_pro_mcp/ida_mcp.py
+++ b/src/ida_pro_mcp/ida_mcp.py
@@ -147,13 +147,18 @@ class MCP(idaapi.plugin_t):
     DEFAULT_PORT = 13337
 
     def init(self):
+        import os
+
         hotkey = MCP.wanted_hotkey.replace("-", "+")
         if __import__("sys").platform == "darwin":
             hotkey = hotkey.replace("Alt", "Option")
 
         self.mcp: "ida_mcp.rpc.McpServer | None" = None
-        self.host = self.DEFAULT_HOST
-        self.port = self.DEFAULT_PORT
+        self.host = os.environ.get("IDA_MCP_HOST", self.DEFAULT_HOST)
+        try:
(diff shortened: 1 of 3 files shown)


mrexodia:
> Add IDA_MCP_HOST / IDA_MCP_PORT environment variables so users can configure the bind address without editing source code or relying on per-IDB netnodes

You should use `idalib-mcp` if you want to expose a server on the LAN. Running ida directly and exposing the MCP in a 'headless' way will be removed as a feature soon.

PR #422, opened 2026-05-19
Restore open_file tool (closes #404)

### LINKED ISSUE #404: open_file tool doesn't exist anymore
Why did you remove the "open_file" tool?
I always use the GUI and the open_file tool was useful

NeKroFR:
Restores the `open_file` MCP tool that was removed in 8461a5f ("Clean up legacy tools"). Originally added in d941072.

Also folds in the Linux executable-path fix from fa86976 (`_get_ida_executable` uses `/proc/self/exe` instead of `sys.executable`).


mrexodia:
Thanks a lot!

PR #437, opened 2026-05-30
sigmaker: vendor v1.8.0 engine (fixes #398)

### LINKED ISSUE #398: `make_signature_for_function` : Full image bytearray is rebuilt on every uniqueness check
This line
https://github.com/mrexodia/ida-pro-mcp/blob/8461a5f7fb249d3fa3a59a4050c678708aaaa199/src/ida_pro_mcp/ida_mcp/_sigmaker.py#L1000 
calls  `InMemoryBuffer.load(mode=SEGMENTS)` inside `_find_all_simd. _load_segments` iterates every segment and does `idaapi.get_bytes(...)` + `bytearray.extend(...)`. For a 5 MB image that's a fresh 5 MB allocation + copy per is_unique call, per instruction added to the growing signature.

A typical function needs 5–40 instructions worth of growth before becoming unique and we do tens of full image rebuilds per signature.
I would like to work on a PR to fix this.

rm-NoobInCoding:
## Summary

Syncs the vendored signature engine (`src/ida_pro_mcp/ida_mcp/_sigmaker.py`) from upstream [ida-sigmaker](https://github.com/mahmoudimus/ida-sigmaker) v1.6.0 → v1.8.0, and fixes #398 (the full image bytearray being rebuilt on every uniqueness check) without adding any third-party dependency.

The engine remains dependency-free (idaapi/idc + stdlib only). The interactive layers (IDA Forms, plugin class, wait-box dialogs, clipboard) are intentionally **not** vendored — this engine runs inside the MCP server, where modal UI is never appropriate.

## Fix for #398 (full image rebuilt on every uniqueness check)

- **`UniqueSignatureGenerator` seeds the candidate set once per `generate()`** and refines it in memory as the pattern grows, instead of re-scanning / rebuilding the image per appended instruction. On the compiled-SIMD path this eliminates the repeated `InMemoryBuffer.load()`.
- **`SignatureSearcher.is_unique` bails at the second match** instead of enumerating every match. For the native `bin_search` fallback (no compiled `_speedups`), this is what keeps uniqueness checks cheap — a short, common prefix no longer enumerates millions of matches. Adds `count_matches` / `find_all_offsets`.

## Other engine updates ported from v1.8.0

- **`WildcardPolicy.for_x86` no longer wildcards immediates.** Literals baked into the encoding don't move between builds, so wildcarding them only removed bytes that would have made the signature unique. More specific, more stable signatures.
- **`GeneratedSignature` orders by `(length, wildcard_count)`** so xref ranking prefers the most specific signature among equal-length candidates.
- **New `MinimalFunctionSignatureGenerator`** finds the shortest unique signature anywhere in a function body (not just from the entry point); exposed via `SignatureMaker.make_function_signature`.
- SIMD seed-and-refine helpers carried over for parity (inert without a compiled `_speedups` module).

Fixes #398.

### DIFF
--- a/src/ida_pro_mcp/ida_mcp/_sigmaker.py
+++ b/src/ida_pro_mcp/ida_mcp/_sigmaker.py
@@ -23,15 +23,10 @@ v1.6.0 vendoring:
     anywhere inside a function body (not just from its entry point).
   - SIMD seed-and-refine helpers are carried over for parity; they are
     inert unless a compiled `_speedups` module is present.
-  - numpy-backed vectorized scanner with a cached, segment-gap-aware image
-    (_NumpyImage / numpy_find_all). When no compiled `_speedups` exists but
-    numpy is importable, signature scanning materializes the image once per
-    session and refines candidate offsets in memory instead of re-running
-    idaapi.bin_search per uniqueness check. On a 370 MB image this measured
-    ~0.13 s per signature warm vs ~1.5-2.2 s for native bin_search (the
-    one-time image build is ~15-17 s, dominated by idaapi.get_bytes). The
-    cache is invalidated on segment-layout change and by
-    invalidate_image_cache(), which the byte-patch tools call.
+
+This file has no third-party dependencies: it relies only on idaapi/idc and
+the standard library. The compiled `_speedups` module is optional; when it is
+absent, scanning falls back to idaapi.bin_search.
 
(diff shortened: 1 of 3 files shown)


mrexodia:
Remove the numpy stuff I do not want any dependencies.

rm-NoobInCoding:
Done

PR #447, opened 2026-06-10
Remove unnecessary API version check for IDA 9.0

rD227:
The _check_required_apis() function blocks IDA 9.0

 but the code already provides complete fallbacks via
hasattr() checks in get_func_name(), get_func_prototype(), and
tinfo_get_udm(). Remove the hard check to allow it to work.


src/ida_pro_mcp/ida_mcp/compat.py
```diff
@@ -84,7 +84,8 @@ def _check_required_apis(version: tuple[int, int, int]) -> None:
     IDA_VERSION: tuple[int, int, int] = cast(tuple[int, int, int], (9, 2, 0))
 else:
     IDA_VERSION = _parse_kernel_version(idaapi.get_kernel_version())
-    _check_required_apis(IDA_VERSION)
+    
+    #_check_required_apis(IDA_VERSION)
```
mrexodia: remove it properly, don't comment it out

PR #456, opened 2026-06-19
Security: Unsafe Stack Variable Deletion Without Proper Validation

tuanaiseo:
## Problem

The delete_stack function in api_stack.py deletes stack frame members based on user-provided name and address. While it checks for special frame members, the validation relies on IDA's internal APIs which may have edge cases. More critically, the declare_stack function allows creating arbitrary stack variables with user-controlled names, offsets, and types, which could corrupt function frames or be exploited for memory corruption in IDA's type system.

**Severity**: `high`
**File**: `src/ida_pro_mcp/ida_mcp/api_stack.py`

## Solution

Add stricter validation for stack variable offsets and sizes. Ensure type names are validated against a whitelist. Consider requiring additional confirmation for destructive operations like delete_stack that modify binary analysis state.

## Changes

- `src/ida_pro_mcp/ida_mcp/api_stack.py` (modified)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced

### DIFF
--- a/src/ida_pro_mcp/ida_mcp/api_stack.py
+++ b/src/ida_pro_mcp/ida_mcp/api_stack.py
@@ -92,7 +92,19 @@ def declare_stack(
                 )
                 continue
 
+            if not var_name or not isinstance(var_name, str):
+                results.append(
+                    {"addr": fn_addr, "name": var_name, "error": "Invalid name"}
+                )
+                continue
+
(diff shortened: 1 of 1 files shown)


mrexodia:
What is an actual input that triggers a crash or exploit in IDA that these checks guard against?

PR #461, opened 2026-06-24
fix(idalib): don't reap busy headless workers on health-probe timeout

lich0821:
### Problem

`IdalibSupervisor.resolve_session()` runs a health probe (TCP connect +
JSON-RPC `ping`) before forwarding every tool call. On the first probe failure
it immediately called `_terminate_worker()` and raised `"… is not reachable"`.

But a headless idalib worker is **single-threaded**. While it is busy running
auto-analysis, building caches, or decompiling, it physically cannot service
the ping until it yields. The supervisor interpreted that silence as death and
killed a perfectly healthy worker, destroying the session.

This is easy to hit with the warmup flags enabled on a large database: open
with `run_auto_analysis=true` (the default), and the very analysis the open
requested keeps the worker busy long enough that the next probe reaps it.

### Fix

Treat "busy" and "dead" as different states. The OS process liveness
(`session.is_alive()`), not a single ping, is the source of truth:

| State | Detection | Action |
|---|---|---|
| Dead | process exited | unregister + reap, raise `not reachable` |
| Busy | alive, ping fails, within grace | keep, retry w/ backoff, raise retryable error |
| Wedged | alive, ping keeps failing past grace | unregister + reap, raise `wedged` |

A grace window (`IDA_MCP_WEDGED_GRACE_SEC`, default 300s) ensures a genuinely
stuck worker (e.g. an analysis loop) still can't pin a worker slot forever —
so the limited worker pool can't be exhausted by zombies. Set it to `0` to
never reap a live worker (pure keep-forever, for single-user long-lived setups).

`_prune_dead_worker_sessions_locked()` is given the same busy-vs-dead
distinction: it now prunes only processes that have actually exited.

### New / changed knobs (all env-overridable, backward-compatible defaults)

| Env | Default | Was |
|---|---|---|
| `IDA_MCP_HEALTH_TCP_TIMEOUT` | `2.0` | hard-coded `0.5` |
| `IDA_MCP_HEALTH_RPC_TIMEOUT` | `10.0` | hard-coded `2.0` |
| `IDA_MCP_HEALTH_RETRIES` | `3` | (none — reaped on first failure) |
| `IDA_MCP_HEALTH_RETRY_BACKOFF` | `1.0` | — |
| `IDA_MCP_WEDGED_GRACE_SEC` | `300` | — |

Plus: `idb_open(idle_ttl_sec=0)` (or any `<= 0`) now means "never self-exit",
for workers a user keeps attached to across many sessions. Positive values keep
the existing MIN-clamp + load-time-padding behavior.

### Compatibility

No default behavior changes for healthy workers — they answer the probe on the
first try and take the unchanged fast path. The only behavioral change is for an
**unreachable** worker: previously reaped on the first failed ping, now retried
and only reaped if its process has exited or it stays wedged past the grace
window. The widened timeouts only ever add latency on the unreachable path.

### Known limitation (pre-existing, out of scope here)

While a worker is opening a very large database, `open_session()` blocks on the
synchronous `call_worker_tool(worker, "idb_open", …)` until the open completes.
Because the HTTP transport services requests on a single thread, unrelated calls
(`tools/list`, `idb_list`, even health probes for *other* sessions) queue behind
that open and appear to hang until it finishes. This is independent of the change
here — the fix operates on the health-probe *decision*, but during a blocking
open requests never reach that decision point.

This PR does not address that serialization; it is called out only for honesty.
In practice the fix's value shows precisely here: a client that retries (and
times out) repeatedly against a worker busy opening a large DB no longer gets
that worker reaped out from under it — the session survives the whole open. A
follow-up could make long opens non-blocking, but that is a larger transport
change orthogonal to busy-worker reaping.

### Tests

`tests/test_idalib_supervisor.py` and `tests/test_worker_lifecycle.py`:

- renamed `…removes_unreachable_worker` → `…removes_dead_worker` (dead = exited)
- new: `keeps_busy_but_alive_worker`, `retries_then_recovers_busy_worker`,
  `reaps_wedged_worker_after_grace`, `grace_zero_never_reaps_live_worker`,
  `recovery_clears_wedged_tracking`
- lifecycle: `zero_or_negative_means_never_exit`,
  `never_exit_overrides_load_time`, `clamps_small_positive_*`,
  `check_never_fires_when_ttl_disabled`
- updated `prunes_unreachable_existing_mapping` to use a dead (exited) process

`pytest tests/` → **198 passed, 114 subtests** (Python 3.14, pytest 9.1.1).

### DIFF
--- a/src/ida_pro_mcp/idalib_server.py
+++ b/src/ida_pro_mcp/idalib_server.py
@@ -100,7 +100,8 @@ def idb_open(
     init_hexrays: Annotated[bool, "Initialize Hex-Rays decompiler after open"] = True,
     idle_ttl_sec: Annotated[
         int,
-        "Minimum idle TTL in seconds before the headless worker self-exits.",
+        "Minimum idle TTL in seconds before the headless worker self-exits. "
+        "Use 0 (or any value <= 0) to keep the worker resident forever.",
     ] = 600,
     preferred_session_id: Annotated[
         str,
--- a/src/ida_pro_mcp/idalib_supervisor.py
+++ b/src/ida_pro_mcp/idalib_supervisor.py
@@ -53,8 +53,26 @@ IDB_MANAGEMENT_TOOLS = {
     "idb_open",
     "idb_list",
 }
-WORKER_TCP_HEALTH_TIMEOUT_SEC = 0.5
-WORKER_RPC_HEALTH_TIMEOUT_SEC = 2.0
+# Health-probe timeouts. A busy worker (mid auto-analysis / decompile) runs
+# single-threaded and cannot answer a JSON-RPC ping until it yields, so a tight
+# RPC timeout misclassifies "busy" as "dead" and gets the worker reaped. These
+# default generously and are env-overridable for slow hosts / huge databases.
+WORKER_TCP_HEALTH_TIMEOUT_SEC = float(os.environ.get("IDA_MCP_HEALTH_TCP_TIMEOUT", "2.0"))
+WORKER_RPC_HEALTH_TIMEOUT_SEC = float(os.environ.get("IDA_MCP_HEALTH_RPC_TIMEOUT", "10.0"))
+
+# When a health probe fails but the worker process is still alive, retry instead
+# of reaping it immediately: a busy worker is not a dead worker. Only a worker
+# whose OS process has actually exited is treated as unreachable.
+WORKER_HEALTH_RETRIES = int(os.environ.get("IDA_MCP_HEALTH_RETRIES", "3"))
+WORKER_HEALTH_RETRY_BACKOFF_SEC = float(os.environ.get("IDA_MCP_HEALTH_RETRY_BACKOFF", "1.0"))
+
+# Grace window (seconds) before a live-but-unresponsive worker is declared
+# wedged and reaped. A worker that keeps failing health probes for longer than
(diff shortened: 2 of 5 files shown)


mrexodia:
Why was it closed?

lich0821:
Thanks for reopening it. I closed the PR because, after using the change for a while, I found that it introduced an incomplete lifecycle behavior rather than because the original bug was invalid.

The original issue still exists: a headless worker that is busy with auto-analysis, cache building, or decompilation can fail the health probe even though the process is still alive, and it should not be reaped immediately.

However, with `idle_ttl_sec <= 0`, the worker can now remain alive indefinitely, while there is currently no explicit close tool for releasing the database/session. This caused databases and workers to stay open longer than intended. I also ran into performance and request-serialization problems when opening large databases, so I continued experimenting with a broader implementation in my fork.

I closed this PR because I did not want the current change to be merged as a complete solution without addressing the missing close/lifecycle behavior.

I’m happy to proceed in whichever way you prefer:
1. keep this PR narrowly focused on preventing busy workers from being reaped, and submit the explicit close/lifecycle handling separately; or
2. update this PR to include the minimum lifecycle changes you consider necessary.

My preference would be to keep the busy-worker fix, explicit close support, and large-database performance work as separate PRs so they remain easier to review.

mrexodia:
That’s alright, just wanted to know the reasoning!

PR #470, opened 2026-07-01
fix: avoid trace flush during idb save to prevent freeze

Creesic:
## Summary

Prevents the trace `savebase` hook from flushing buffered trace records while IDA is saving the database.

## Root Cause

`idb_save` enters IDA's save path, which triggers `IDB_Hooks.savebase`. The trace hook previously called `flush()` there, causing reentrant netnode writes during database save. In GUI IDA this can destabilize the save path and crash.

## Impact

Trace records still flush through normal thresholds, trace iteration, shutdown, and close. They are no longer written from inside the save hook.

## Validation

- Regression failed before the fix
- `uv run ida-mcp-test tests/crackme03.elf -p "*savebase_does_not_flush*" -q` -> 1 passed
- `uv run ida-mcp-test tests/crackme03.elf -c trace -q` -> 12 passed
- Manual test in IDA confirmed `idb_save` no longer crashes.

### DIFF
--- a/src/ida_pro_mcp/ida_mcp/tests/test_trace.py
+++ b/src/ida_pro_mcp/ida_mcp/tests/test_trace.py
@@ -197,6 +197,38 @@ def test_trace_idb_shutdown_flushes_pending_buffer():
 
 
 @test()
+def test_trace_idb_savebase_does_not_flush_pending_buffer():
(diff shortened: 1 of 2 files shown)


mrexodia:
Cheers!

PR #478, opened 2026-07-13
Fix stale strings cache when switching binaries in idalib worker

### LINKED ISSUE #476: The strings cache is process-global, not per-session and gets stale after switching binaries in the idalib worker
`_strings_cache` in `ida_mcp/api_core.py` is a single module-level global, built on first access. The function `invalidate_strings_cache()` exists but is never called anywhere in the code.

`IDASessionManager` (idalib_session_manager.py) supports multiple binaries open in one idalib process, switching the active database via `idapro.close_database()`/`open_database()` in `_activate_database_path()` (called from both `open_binary()` and `activate_session()`), but none of these paths touch `_strings_cache`. Once built, the cache never refreshes for the lifetime of the worker process, no matter how many sessions are opened/switched afterward.

How to reproduce:
1. `idb_open(A)` -> builds `_strings_cache` from binary A.
2. `idb_open(B)` (different binary, same worker).
3. `search_text(kind="strings")` / `survey_binary` still return binary A's  strings (and addresses), because `_strings_cache is not None`, preventing the rebuild.

Suggested fix: key the cache by session_id (`dict[str, list[tuple[int, str]]]`) instead of a single list, and drop the entry in `close_session()`. Avoids rebuilding strings when switching back to an already-open session.

Alternative: call `invalidate_strings_cache()` inside `_activate_database_path()` whenever the active database changes.

NeKroFR:
Fixes #476

`_strings_cache` in `api_core.py` is process-global and was built once, then never refreshed. After the idalib worker switched databases, strings queries kept returning the previous binary's strings.

Added a `_close_active_database()` helper in `IDASessionManager` that drops the cache on every DB teardown/switch (`open_binary`, `activate_session`, `close_session`, `close_all_sessions`). Closing an inactive session leaves the active binary's cache intact.

**Why not the per-session dict:** `api_core` is session-agnostic. Its tools run against whatever DB idalib currently has open and never receive a `session_id`. Keying the cache by session would force `api_core` to ask the session manager
which session is active, inverting the dependency. Invalidating on switch keeps the coupling one-way (manager -> api_core).
The only cost is a lazy rebuild on switch-back, cheap next to the reopen + auto-analysis a switch already does.

### DIFF
--- a/src/ida_pro_mcp/idalib_session_manager.py
+++ b/src/ida_pro_mcp/idalib_session_manager.py
@@ -16,6 +16,8 @@ from datetime import datetime
 import idapro
 import ida_auto
 
+from .ida_mcp.api_core import invalidate_strings_cache
+
 logger = logging.getLogger(__name__)
 
 
@@ -115,8 +117,7 @@ class IDASessionManager:
             logger.info(f"Closing session: {session_id} ({session.input_path.name})")
 
             if self._active_session_id == session_id:
-                idapro.close_database()
-                self._active_session_id = None
+                self._close_active_database()
 
             del self._sessions[session_id]
             logger.info(f"Session closed: {session_id}")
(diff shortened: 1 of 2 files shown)


NeKroFR:
@mrexodia looks good to you or do you want me to change something ?

mrexodia:
I'm cool with it, was waiting for CI

PR #481, opened 2026-07-17
feat(installer): add Kimi Code plugin

s0kil:
Kimi Code reads MCP servers from `<KIMI_CODE_HOME>/mcp.json`, falling back
to `~/.kimi-code/mcp.json`, and from `.kimi-code/mcp.json` for project scope.
Both use the standard `{"mcpServers": {...}}` shape, so no special JSON
structure is needed.


mrexodia:
Thanks!

PR #488, opened 2026-07-24
fix(idalib): avoid SIGTERM/SIGINT handler deadlock in idalib_server

### LINKED ISSUE #487: idalib_server deadlocks on SIGTERM/SIGINT; detached workers leak as unkillable PPID=1 orphans
## Summary

`idalib_server` workers deadlock permanently when they receive SIGTERM/SIGINT. Because spawned workers are deliberately detached from the supervisor (`start_new_session=True`), every wedged worker survives as a PPID=1 orphan that ignores further SIGTERMs and can only be killed with SIGKILL. On a machine where an MCP client (e.g. Claude Code) starts the `idalib-mcp` supervisor once per session, this leaks roughly one orphaned idalib process per session. In my case they accumulated to **~774 orphaned processes eating ~28 GB of swap** before I noticed.

## Environment

- macOS 26.5 (arm64), Python 3.11.15 (Homebrew), IDA Professional 9.3 with idalib
- ida-pro-mcp 2.0.0 (installed as a Claude Code plugin; supervisor runs as `idalib-mcp --stdio`)

## Root cause

The worker runs a **single-threaded** HTTP server on the **main thread**:

```python
# idalib_server.py main()
MCP_SERVER.serve(host=..., port=..., background=False, ...)
```

With `background=False`, `zeromcp.McpServer.serve()` uses plain `HTTPServer` and calls `serve_forever()` directly on the calling (main) thread.

The signal handler then does:

```python
def cleanup_and_exit(signum, frame):
    logger.info("Signal %s received; shutting down", signum)
    try:
        MCP_SERVER.stop()
    except Exception:
        ...
```

`MCP_SERVER.stop()` calls `self._http_server.shutdown()`, which per stdlib docs *blocks until `serve_forever()` returns* and **must be called from a different thread than the one running `serve_forever()`** (the zeromcp source even says so in a comment). But a Python signal handler runs on the main thread — on top of the very frame that is parked inside `serve_forever()`. The handler waits forever for a loop that is frozen underneath it. Classic self-deadlock.

It gets worse: `stop()` sets `self._running = False` *before* blocking, and `stop()` starts with `if not self._running: return`. So after the first (deadlocked) call:

- the **idle-TTL watchdog** (`WorkerLifecycle`, 600 s) fires later, calls `MCP_SERVER.stop()` from its own thread — and returns immediately as a **no-op**;
- **repeated SIGTERMs** re-enter the handler, call `stop()` — also a no-op.

The process is permanently wedged; only SIGKILL works.

## Evidence

`sample` (macOS profiler) of a wedged orphan (PPID=1, all future SIGTERMs ignored):

```
Main thread:
  serve_forever -> select_poll_poll
    -> _PyErr_CheckSignalsTstate            # signal arrived
      -> (cleanup_and_exit -> MCP_SERVER.stop -> HTTPServer.shutdown)
        -> lock_PyThread_acquire_lock -> acquire_timed
          -> __psynch_cvwait                # waits forever for serve_forever to exit
```

## How the orphans are born (typical MCP client lifecycle)

1. Client session starts → supervisor `idalib-mcp --stdio` starts.
2. First `tools/list` → supervisor spawns a schema worker (`_schema_or_idle_worker`), detached via `start_new_session=True`.
3. Session ends → supervisor's `shutdown()` calls `_terminate_worker()` → sends the schema worker **SIGTERM** → worker deadlocks as above.
4. `_terminate_worker` would SIGKILL it after `proc.wait(timeout=10)`, but the client usually kills the supervisor process well before those 10 s elapse → the fallback never runs.
5. The wedged worker reparents to PID 1 and lives forever. One orphan per session.

(Workers that never receive a signal are fine — the idle-TTL self-exit works. Only the signal path wedges them, and it also disarms the TTL path afterwards.)

## Reproduction (no MCP client needed)

```python
# spawn a worker, wait until it answers JSON-RPC ping on /mcp, then:
proc.send_signal(signal.SIGTERM)
proc.wait(timeout=10)   # -> TimeoutExpired; process is deadlocked forever
```

Before the fix: the worker never exits (10 s, 10 min — doesn't matter). After the fix: exits with rc=0 within milliseconds.

## Suggested fix

Run `MCP_SERVER.stop()` from a short-lived helper thread inside the signal handler, which is exactly the "different thread" that `HTTPServer.shutdown()` requires. The main thread then leaves `serve_forever()` normally and `main()`'s `finally` block performs the usual cleanup (`_LIFECYCLE.stop()`, discovery deregistration, `close_all_sessions()`).

PR incoming.

Zzoomrus:
Fixes #487.

## Problem

`idalib_server` runs a single-threaded `HTTPServer` on the main thread (`MCP_SERVER.serve(..., background=False)`). Its SIGTERM/SIGINT handler called `MCP_SERVER.stop()` directly — but `stop()` blocks in `HTTPServer.shutdown()`, which waits for `serve_forever()` to return and must be called from a *different* thread than the one running it. The handler runs on the main thread, on top of the frame parked inside `serve_forever()`, so the process deadlocks forever.

Since `stop()` flips `_running = False` before blocking, every subsequent `stop()` call (the idle-TTL watchdog, repeated signals) becomes a no-op. Combined with `start_new_session=True` in the supervisor's `_spawn_worker()`, each wedged worker survives as a PPID=1 orphan that only SIGKILL can remove — leaking roughly one idalib process per supervisor lifetime (I had ~774 of them holding ~28 GB of swap). Full analysis with a stack sample of a wedged process is in #487.

## Fix

Run `MCP_SERVER.stop()` on a short-lived daemon thread from the signal handler. That is exactly the "different thread" `shutdown()` requires; the main thread then exits `serve_forever()` normally and `main()`'s `finally` block performs the usual cleanup (`_LIFECYCLE.stop()`, discovery deregistration, `close_all_sessions()`).

## Verification

On macOS 26.5 / Python 3.11 / IDA Professional 9.3 (idalib):

- **Before**: spawn a worker, wait until it answers JSON-RPC `ping` on `/mcp`, send SIGTERM → still alive 10 s later (and indefinitely; `sample` shows the handler blocked in `HTTPServer.shutdown()` under `serve_forever()`).
- **After**: same scenario → clean exit with rc=0 within milliseconds.
- Idle-TTL self-exit path still works after the change (worker with a shortened TTL and no requests exits on its own).
- `pytest tests/test_worker_lifecycle.py tests/test_idalib_supervisor.py` → 50 passed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/ida_pro_mcp/idalib_server.py
+++ b/src/ida_pro_mcp/idalib_server.py
@@ -2,6 +2,7 @@ import argparse
 import logging
 import os
 import signal
+import threading
 import time
 from pathlib import Path
 from typing import Annotated, Any, TypedDict
@@ -248,10 +249,19 @@ def main():
 
     def cleanup_and_exit(signum, frame):
         logger.info("Signal %s received; shutting down", signum)
-        try:
-            MCP_SERVER.stop()
-        except Exception:
-            logger.exception("MCP_SERVER.stop() failed in signal handler")
+        # MCP_SERVER.stop() blocks until serve_forever() returns, but this
+        # handler runs on the main thread — the same thread that is parked
+        # inside serve_forever() underneath this frame. Calling stop() here
+        # deadlocks the process forever (and, having flipped _running to
+        # False, turns every later stop() — watchdog TTL, repeat signals —
+        # into a no-op). Stop from a helper thread instead.
+        def _stop():
+            try:
+                MCP_SERVER.stop()
+            except Exception:
+                logger.exception("MCP_SERVER.stop() failed in signal handler")
+
+        threading.Thread(target=_stop, daemon=True, name="signal-stop").start()
 
     signal.signal(signal.SIGINT, cleanup_and_exit)
     signal.signal(signal.SIGTERM, cleanup_and_exit)


mrexodia:
thanks!

PR #491, opened 2026-07-25
Add Dockerfile for idalib-mcp HTTP mode

nrf24l01:
Add Dockerfile for idalib-mcp HTTP mode


mrexodia:
I'm fine adding the `/health` endpoint, but I don't really want to merge the Dockerfile setup since that is more personal to your situation.

PR #493, opened 2026-07-26
Add Python 3.8 / IDA 7.7 compatibility

wxx6:
- Replace match/case with if/elif/else (Python 3.10+ only)
- Move NotRequired, Annotated, Required, is_typeddict, TypeAlias from typing to typing_extensions
- Add from __future__ import annotations to files missing it
- Replace runtime list[X]/dict[X]/tuple[X]/set[X] with List[X]/Dict[X]/Tuple[X]/Set[X] for Python 3.8
- Fix UnionType import (try/except for Python <3.10)
- Fix functools.cache -> lru_cache(maxsize=None)
- Fix dataclass(slots=True) -> remove slots (Python 3.10+ only)
- Fix X | Y union syntax in functional TypedDict bodies
- Adjust pyproject.toml: requires-python >=3.8, remove idapro dep
- Add typing_extensions dependency via pip install

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,7 +3,7 @@ name = "ida-pro-mcp"
 version = "2.0.0"
 description = "Vibe reversing with IDA Pro"
 readme = "README.md"
-requires-python = ">=3.11"
+requires-python = ">=3.8"
 authors = [
     { name = "mrexodia" },
     { name = "can1357" },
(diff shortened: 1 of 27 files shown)


mrexodia:
Why would you use 7.7? Even if you pirate IDA you can get the latest version...

wxx6:
> Why would you use 7.7? Even if you pirate IDA you can get the latest version...

Just used to it. Personal preference.

PR #495, opened 2026-07-26
feat(idalib): add idb_close tool to save and release worker sessions

### LINKED ISSUE #486: No way to close an IDALib session after idalib_close was removed
Hi, I noticed that `idalib_close` was removed in commit `22aae5f` as part of
the persistent worker refactor.

The current MCP interface provides `idb_open` and `idb_list`, but there is no
corresponding `idb_close`. The internal code still has
`IDASessionManager.close_session()` and `idapro.close_database()`, but they
are not exposed as MCP tools.

This becomes a problem when analyzing multiple files:

1. Start the supervisor with the default `--max-workers 4`.
2. Open four different databases with `idb_open`.
3. Finish analyzing one of them.
4. Try to open a fifth database.
5. The request fails with:

Maximum idalib worker count reached (4).
Wait for an existing worker to be released or increase --max-workers.

There is currently no way for the client to explicitly release a worker.
Waiting for the idle TTL is inconvenient for agent workflows, especially
because agents normally call `idb_list` to inspect available sessions. The
health probes performed while listing sessions may also keep workers active
longer.

Restarting the whole supervisor releases the workers, but it also closes every
other active database. Increasing `max-workers` only postpones the same issue
and uses more memory.

Was removing `idalib_close` intentional? If so, what is the intended workflow
for long-running or cross-file analysis once the worker limit is reached?

Would you consider adding a supervisor-level tool such as:

```python
idb_close(database: str, save: bool = True)

For supervisor-owned headless workers, it could save the database, unregister
the session and terminate the worker. For GUI sessions or externally adopted
workers, it could detach without terminating the external process.

The old implementation already handled most of the owned-worker cleanup:

- removed the session from the supervisor registry;
- removed path and context bindings;
- called the worker-side close tool;
- terminated the owned worker.

The new persistent/adopted worker behavior would need some additional
handling, but having an explicit release operation would make multi-file
analysis much more practical.

NeKroFR:
Fixes #486. The idalib supervisor caps concurrent workers at `--max-workers` (default 4) and workers only self-exit after their idle TTL, so opening more binaries fails with *"Maximum idalib worker count reached"* with no way to release a session.

Adds an `idb_close(database, save=True)` management tool that saves the DB (via the worker's `idb_save`), unregisters the session and its path bindings, and terminates the owned worker to free a slot. Adopted GUI/worker instances are detached, not killed.

If you see anything to change or ways to improve the feature, feel free to let me know. Happy to iterate.

### DIFF
--- a/README.md
+++ b/README.md
@@ -194,7 +194,7 @@ _Note_: The `idalib` feature was contributed by [Willi Ballenthin](https://githu
 
 ## Headless idalib Session Model
 
-`idalib-mcp` is a supervisor that keeps each open database in its own idalib worker process. Workers register themselves in a host-local discovery directory and outlive the supervisor that spawned them; any subsequent supervisor that wants the same path adopts the running worker. A worker self-exits when no request has hit it for its idle TTL (default 1 hour). There is no `idb_close` tool — clients that no longer care about a database simply stop using it, and only the user can close a GUI window.
+`idalib-mcp` is a supervisor that keeps each open database in its own idalib worker process. Workers register themselves in a host-local discovery directory and outlive the supervisor that spawned them; any subsequent supervisor that wants the same path adopts the running worker. A worker self-exits when no request has hit it for its idle TTL (default 1 hour). Call `idb_close` to release a worker eagerly (freeing a slot toward `--max-workers`), adopted GUI/worker instances are detached rather than killed.
 
 `idb_open` picks the backend via its `mode` parameter:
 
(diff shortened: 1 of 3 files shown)


NeKroFR:
I think this one needs a review, is it good as is or would you change anything? @mrexodia? I think you should take a look before aproving the PR.

mrexodia:
I believe in your skills, let's push it to production 😉

PR #499, opened 2026-07-30
fix(idb_save): never DBFL_KILL working files of an open database

### LINKED ISSUE #498: `idb_save` uses `DBFL_KILL` in headless mode, corrupting the still-open database on repeated saves within a session
The headless branch of `idb_save` in `api_core.py` calls `save_database` with `DBFL_KILL | DBFL_COMP`:

https://github.com/mrexodia/ida-pro-mcp/blob/main/src/ida_pro_mcp/ida_mcp/api_core.py#L873

```python
else:
    # Headless idalib: safe to pack into a single compressed file.
    flags = ida_loader.DBFL_KILL | ida_loader.DBFL_COMP
    ok = bool(ida_loader.save_database(save_path, flags))
```

`DBFL_KILL` deletes the loose working files (`.id0`/`.id1`/`.id2`/`.nam`/`.til`) after packing them into the compressed file. The inline comment justifies this by asserting that headless idalib "has no live loose files to clobber."

That assumption only holds for a save-then-close pattern. It breaks in the scenario this MCP server is actually designed for: a long-lived headless analysis session where the database stays open and `idb_save` is called multiple times.

## The bug

When the database is open in `idalib`, it is backed by live loose working files, just as it is in the GUI. On the first `idb_save` call, `DBFL_KILL` packs and then deletes those working files out from under the still-open database. Any subsequent analysis, or a later `idb_save` call in the same session, then operates on a database whose backing files have been removed - the exact corruption pattern the GUI branch already guards against per issue #446 

The real variable that determines safety here is whether the database is still open, not whether the environment is GUI vs headless. The current code keys the decision on the wrong condition: the GUI branch is careful because the DB is open, but the headless branch assumes the DB might as well be closed. In a headless MCP session the DB is open the whole time, so headless needs the same care.

## Reproduction (conceptual)

- Start a headless idalib MCP session; the database stays open.
- Call `idb_save` once (working files packed and killed).
- Rename a global variable and then call `idb_save` again.
- The session is now operating on a database whose loose working files were deleted --> No global variable name is saved

## Suggested fix

Don't  use `DBFL_KILL` while the database is open, in either mode. For the normal in-session save, use `DBFL_COMP` alone so the compressed copy is written but the live working files are preserved:

```python
else:
    # Headless idalib: DB is still open, so keep the loose working files.
    # Do NOT DBFL_KILL them out from under the open database.
    ok = bool(ida_loader.save_database(save_path, ida_loader.DBFL_COMP))
```

NeKroFR:
Closes #498.

- `idb_save` no longer branches on GUI vs headless: the database is open either way, so the loose `.id0/.id1/.id2/.nam/.til` files are live and `DBFL_KILL` corrupts them mid-session. In-place saves use IDA's native save, an explicit different destination gets a `DBFL_COMP` copy.
- `close_session` now cleans up the leftover working files after terminating the worker, since the SIGTERM path has no clean `close_database()`.
- New `test_idb_save_keeps_working_files` verified failing on the old behavior, passing on the fix. Full fixture suites green (349 / 323+2 skipped), supervisor tests 69 passed.

### DIFF
--- a/src/ida_pro_mcp/ida_mcp/api_core.py
+++ b/src/ida_pro_mcp/ida_mcp/api_core.py
@@ -835,43 +835,24 @@ def idb_save(
 ) -> IdbSaveResult:
     """Save active IDB to disk, optionally to a provided path.
 
-    In the GUI (idaq) the open database is backed by loose working files
-    (.id0/.id1/.id2/.nam/.til) that IDA actively manages; packing+killing them
-    out from under the running GUI corrupts the database on the next reopen
-    (issue #446). So in GUI mode this uses IDA's native in-place save
-    (equivalent to Ctrl+W / save_database(None, 0)), and for an explicit
-    different destination writes a compressed copy WITHOUT killing the live
-    working files.
-
-    Only headless idalib — which has no live loose files to clobber — packs into
-    a single compressed .i64/.idb, removing the loose working files.
+    The open database is backed by loose working files (.id0/.id1/.id2/.nam/.til)
+    that IDA actively manages, headless included. Killing them out from under the
+    open database corrupts it (issues #446, #498), so DBFL_KILL is never used
+    here: an in-place save uses IDA's native save (Ctrl+W), and an explicit
+    different destination gets a compressed copy while the working files stay.
     """
     try:
         save_path = path.strip() if path else ""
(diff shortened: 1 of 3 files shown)


mrexodia:
Thx!

Issue #68, opened 2025-05-13
ida pro doesn't show plugin


eadrick1:
I am using ida pro version 8.3 and the plugin does not appear

Chapoly1305:
Double check you need to use Python 3.11/3.12. I also had this issue, upgraded Ubuntu to 24.04 and it resloved.

mrexodia:
It also doesn’t work with IDA Free just in case you are using that.

PR #500, opened 2026-07-31
feat: expose ignore_micro (skippable instructions) in disasm output and add set_ignore_micro tool

sonyps5201314:
## Motivation

While using ida-pro-mcp with an AI assistant to analyze a DLL containing hand-written x86 assembly, I encountered a case where the decompiler produced incorrect pseudocode. The root cause turned out to be IDA's **"skippable instructions"** mechanism (`ignore_micro`): the processor module had incorrectly marked two `push` instructions as function **prolog**, causing the decompiler to NOP them during microcode generation.

I reported this issue on the Hex-Rays community forum:
- **[The x86 decompiler plugin incorrectly NOPs necessary assembly instructions, leading to incorrect C pseudocode generation](https://community.hex-rays.com/t/the-x86-decompiler-plugin-incorrectly-nops-necessary-assembly-instructions-leading-to-incorrect-c-pseudocode-generation/728)**

Following the official guidance from [Igor's tip of the week #68: Skippable instructions](https://hex-rays.com/blog/igors-tip-of-the-week-68-skippable-instructions), I was able to fix the issue by manually clearing the incorrect `ignore_micro` flags. However, the entire diagnosis and fix process required deep knowledge of IDA internals (netnodes, `ignore_micro_t`, etc.) — something an AI assistant currently has no visibility into.

**This PR gives the AI full visibility into skippable instruction flags and the ability to fix them**, making it possible for an AI to autonomously diagnose and repair this class of decompilation errors.

### 1. Expose `ignore_micro` status in all disassembly output

Every disassembly output path now indicates when an instruction is marked as skippable:

- **`disasm` tool** (structured output): Adds an `ignore_micro` field (`"prolog"`, `"epilog"`, or `"switch"`) to `DisassemblyLine`
- **`func_profile`** / **`insn_query`** / **`get_assembly_lines`** (text output): Appends `[SKIP:<type>]` suffix to marked instruction lines

This means the AI can see at a glance which instructions the decompiler will skip, regardless of which tool it uses.

### 2. Add `set_ignore_micro` tool

A new batch modification tool that sets or clears `ignore_micro` flags:
```json
{"ops": [
  {"addr": "0x100010C5", "im_type": "none"},
  {"addr": "0x100010C6", "im_type": "prolog"}
]}
```

### 3. Cross-session state tracking with automatic backup/restore

When `set_ignore_micro` modifies a flag for the first time, the original value is saved as a regular comment visible in IDA's GUI:
```
push ebx  ; y [ida-mcp: restore ignore_micro to prolog]
```

This ensures:
- A **new AI session** can read the comment and know exactly what value to restore to
- The backup comment is **automatically removed** when the instruction is restored to its original state
- If the AI sets a wrong value, the result includes a `warning` field with explicit guidance:
  ```json
  {
    "addr": "100010c5",
    "previous": "epilog",
    "current": "none",
    "restore_to": "prolog",
    "warning": "This instruction's original value was 'prolog'. To restore, use set_ignore_micro with im_type='prolog', not 'none'."
  }
  ```

### 4. Self-documenting design

All docstrings (`disasm`, `DisassemblyLine`, `set_ignore_micro`) explain what `ignore_micro` means and how to use it. Even if `set_ignore_micro` is disabled via the config page, the AI can still understand the `ignore_micro` field in disasm output.

## Files Changed

| File | Lines | Summary |
|------|-------|---------|
| `utils.py` | +93 | Constants, helpers (`im_get_tag`, `im_backup_comment`, etc.), `DisassemblyLine.ignore_micro` field |
| `api_analysis.py` | +31 | `disasm`, `_disasm_lines_limited`, `insn_query`, `get_assembly_lines` all output ignore_micro info |
| `api_modify.py` | +118 | New `set_ignore_micro` tool with backup/restore logic |

## Example Workflow

**Before this PR**: The AI sees wrong pseudocode, has no way to know *why*, and cannot fix it.

**After this PR**:
1. AI calls `disasm("__Hello2")` → sees `"ignore_micro": "prolog"` on `push ebx`
2. AI recognizes this instruction shouldn't be skipped → calls `set_ignore_micro(addr="0x100010C5", im_type="none")`
3. AI re-decompiles → correct pseudocode ✅
4. In a future session, another AI sees `[ida-mcp: restore ignore_micro to prolog]` in comments → knows exactly how to undo if needed

### DIFF
--- a/src/ida_pro_mcp/ida_mcp/api_analysis.py
+++ b/src/ida_pro_mcp/ida_mcp/api_analysis.py
@@ -45,6 +45,8 @@ from .utils import (
     InsnPattern,
     FuncProfileQuery,
     AnalyzeBatchQuery,
+    im_get_node,
+    im_get_tag,
 )
 from . import compat
 
@@ -616,13 +618,16 @@ def _limit_items(items: list, limit: int) -> tuple[list, bool]:
 def _disasm_lines_limited(func: ida_funcs.func_t, max_insns: int) -> tuple[list[str], bool]:
     lines: list[str] = []
     truncated = False
+    im_node = im_get_node()
     for item_ea in idautils.FuncItems(func.start_ea):
         if len(lines) >= max_insns:
             truncated = True
             break
         line = ida_lines.generate_disasm_line(item_ea, 0)
         instruction = ida_lines.tag_remove(line) if line else ""
-        lines.append(f"{item_ea:x}  {compact_whitespace(instruction)}")
+        tag = im_get_tag(item_ea, im_node)
+        suffix = f"  [SKIP:{tag}]" if tag else ""
+        lines.append(f"{item_ea:x}  {compact_whitespace(instruction)}{suffix}")
     return lines, truncated
 
 
@@ -795,7 +800,17 @@ def disasm(
         bool, "Compute total instruction count (default: false)"
     ] = False,
 ) -> DisasmResult:
-    """Disassemble function with offset/max_instructions pagination and optional total count."""
+    """Disassemble function with offset/max_instructions pagination and optional total count.
+
+    Each instruction line may include an 'ignore_micro' field ("prolog", "epilog",
+    or "switch") indicating the decompiler will skip/NOP this instruction.
+    If decompiled output looks wrong, an incorrectly marked instruction may be
+    the cause. Use set_ignore_micro(im_type='none') to clear the mark and
+    re-decompile.
+
+    If a comment contains '[ida-mcp: restore ignore_micro to <type>]', it means
+    the instruction was previously modified. To undo, call set_ignore_micro with
+    im_type set to the <type> value shown after 'to' in that comment."""
 
     # Enforce max limit
     if max_instructions <= 0 or max_instructions > 50000:
@@ -833,6 +848,8 @@ def disasm(
         total_count = 0
         more = False
 
+        _im_node = im_get_node()
+
         def _maybe_add(ea: int) -> bool:
             nonlocal seen, total_count, more
             if include_total:
@@ -856,6 +873,9 @@ def disasm(
                 refs = _collect_line_refs(ea)
                 if refs:
                     entry["refs"] = refs
+                tag = im_get_tag(ea, _im_node)
+                if tag:
+                    entry["ignore_micro"] = tag
                 lines.append(entry)
                 seen += 1
                 return True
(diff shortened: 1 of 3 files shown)


sonyps5201314:
**Test DLL Project:**

`dllmain.cpp`
```cpp
// dllmain.cpp : Defines the entry point for the DLL application.
#include "pch.h"

BOOL APIENTRY DllMain( HMODULE hModule,
                       DWORD  ul_reason_for_call,
                       LPVOID lpReserved
                     )
{
    switch (ul_reason_for_call)
    {
    case DLL_PROCESS_ATTACH:
    case DLL_THREAD_ATTACH:
    case DLL_THREAD_DETACH:
    case DLL_PROCESS_DETACH:
        break;
    }
    return TRUE;
}

// Helper that implements the original logic. Keep normal prologue/epilogue.
// Give C linkage so it can be called from assembly (_Hello_impl symbol).
extern "C" int Hello_impl(int x, int y, const char* pszCaption)
{
    int a = x + y;
    char szText[256];
    sprintf(szText, "Hello from DLL! x + y = %d", a);
    return MessageBoxA(NULL, szText, pszCaption, MB_OK);
}

// Naked exported entry that expects parameters in EAX (x), EBX (y), ECX (psz)
// and returns result in ECX.
// Declare Hello2 implemented in assembly.
extern "C" int Hello2();

extern "C" int __declspec(naked) Hello(/*int x, int y, const char* pszCaption*/)
{
    __asm {
        // push arguments for Hello_impl: psz, y, x (right-to-left)
        push ecx
        push edx
        push eax
        call Hello_impl
        // result in EAX; move it into ECX as required by the new convention
        mov ecx, eax
        // clean up the pushed arguments
        add esp, 12
        // return; caller will read ECX for the return value
        ret
    }
}
```

`Hello2.asm`
```asm
.386
.model flat, C
.code

; Call the existing exported Hello (naked) implemented in C++.
; The symbol is _Hello for C linkage.
extrn Hello_impl:PROC

PUBLIC _Hello2
_Hello2 PROC
    ; Caller passes x->EAX, y->EBX, psz->ECX
    ; Push arguments for _Hello_impl (right-to-left): psz, y, x
    push ecx
    push ebx
    push eax
    call Hello_impl
    ; move return (EAX) into ECX as required by new convention
    mov ecx, eax
    ; clean up stack (3 dwords)
    add esp, 12
    ret
_Hello2 ENDP

END
```

sonyps5201314:
**Execution result:**
image
image

**After running a similar prompt in the AI IDE:**
```text
@mcp:ida-pro-mcp Please help me correct the decompilation results for __Hello2.
```
image
image

PR #501, opened 2026-08-04
feat(idalib): configure startup worker idle timeout

amitelka:
Calls to `idb_open` already accept `idle_ttl_sec`, but the positional startup path always requests a 600-second timeout. In long-running coding-agent sessions, more than ten minutes can pass between IDA calls while the agent analyzes earlier results or works in the surrounding codebase. Keeping the initialized worker available currently requires periodic requests.

Add CLI and environment controls while preserving the existing default. CLI values take precedence over `IDA_MCP_IDLE_TIMEOUT`, which takes precedence over the 600-second default. Zero disables idle expiry.

## Verification

- `uv run pytest -q tests` (`251 passed, 114 subtests passed`)
- Live idle A/B with no worker RPC traffic:
  - `--idle-timeout 10`: registration and listening socket were gone after 15 seconds.
  - `--idle-timeout 0`: registration, listening socket, worker, and supervisor were still present after 630 seconds.

### DIFF
--- a/README.md
+++ b/README.md
@@ -210,7 +210,7 @@ _Note_: The `idalib` feature was contributed by [Willi Ballenthin](https://githu
 
 ## Headless idalib Session Model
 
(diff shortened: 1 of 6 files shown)


amitelka:
Rebased onto current main after #505 merged, the PR now is scoped to just the startup idle timeout configuration

PR #502, opened 2026-08-05
docs: add Linux idalib activation example

Fuwaki:
The IDALib activation instructions currently include Windows and macOS paths, but no Linux equivalent. This adds the matching Linux command with a generic IDA installation path.

I verified the Linux activation flow with an IDA Pro 9.0.240925 portable installation. The bundled activation script configured IDALib successfully, and `idalib-mcp` initialized IDALib and Hex-Rays, opened and analyzed an ELF binary, and returned a healthy server status.


mrexodia:
FYI since IDA 9.4 the installer does this automatically.

PR #503, opened 2026-08-05
feat: py_eval: add opt-in persistent locals via new_locals parameter

huzpsb:
## Problem

`py_eval` documents its trailing-expression return as "Jupyter-style". This wording is misleading: AI assistants (and users) reasonably assume Jupyter-style implies a **persistent namespace**, and repeatedly write code that references variables from earlier `py_eval` calls — which silently fails, because the previous implementation rebuilt both globals and locals from scratch on every call.

## Solution

Add an opt-in persistence mechanism:

- New parameter `new_locals: bool = True`
  - `True` (default): fresh locals namespace per call — **behavior is unchanged** from before
  - `False`: reuse the module-level stored locals dict, so variables, imports, and definitions from previous calls remain visible (actual Jupyter-like persistence)
- Passing `new_locals=True` again resets the stored namespace
- Persisted locals are also merged into the per-call globals so they resolve correctly in the single-expression `eval` fast path

The globals namespace (IDA modules, helper functions) is still rebuilt per call; only locals persist.

The docstring and parameter description now explicitly state this behavior, removing the ambiguity that "Jupyter-style" previously created.

### DIFF
--- a/src/ida_pro_mcp/ida_mcp/api_python.py
+++ b/src/ida_pro_mcp/ida_mcp/api_python.py
@@ -102,6 +102,9 @@ class PythonExecResult(TypedDict):
     stderr: str
 
 
+_persisted_locals: dict = {}
+
+
 # ============================================================================
 # Python Evaluation
 # ============================================================================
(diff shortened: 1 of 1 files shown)


mrexodia:
I think it's better to make this the default behavior and always persist them so the behavior matches the documentation.

huzpsb:
> I think it's better to make this the default behavior and always persist them so the behavior matches the documentation.

Agreed — flipped the default, locals now persist by default.

mrexodia:
cheers!

PR #505, opened 2026-08-08
fix(idalib): keep busy workers reachable instead of reaping them

### LINKED ISSUE #494: idalib worker becomes unreachable after successful analyze_function on a large .i64
### Summary

An `idalib-mcp` headless worker becomes unreachable after a successful Hex-Rays-backed query on a large existing `.i64` database. The next lightweight `server_health` call never returns, then the client receives `Worker for session ... is not reachable`.

### Environment

- Windows 11
- IDA Pro 9.3 with Hex-Rays ARM64 available
- `ida-pro-mcp` local project version: `2.0.0`
- Local source commit: `060d15d33c4567c6179d002fe1c6805270e2738a` (2026-07-24)
- Transport: Codex desktop using `uv run idalib-mcp --stdio`
- Target: a pre-analyzed ARM64 Unity IL2CPP `libil2cpp.so` `.i64` database

### Minimal reproduction

1. Open the existing database:

```text
idb_open(input_path=<existing .i64>, mode=force_headless, run_auto_analysis=false, build_caches=false, init_hexrays=true)
```

This succeeds. `auto_analysis_ready=true` and `hexrays_ready=true`.

2. Run one read-only analysis:

```text
analyze_function(database=<session>, addr=0x28CCF40, include_asm=false)
```

This succeeds quickly (about 0.7 s) and returns truncated pseudocode.

3. Immediately run:

```text
server_health(database=<same session>)
```

The request hangs and ultimately reports:

```text
Worker for session <session> is not reachable
```

A direct `decompile` for the same function also exceeds 100 seconds and ends with the same worker-unreachable error.

### Expected behavior

A busy worker should remain registered. `server_health` should return a structured `busy`/timeout result, or the original analysis should return a structured timeout, without losing the session. A later small read-only query should still be possible.

### Notes

This looks related to open PRs #483 (do not reap busy workers on probe timeout) and #485 (thread the idalib HTTP layer / expose busy state). The local source predates those changes. This report is separate because the sequence was reproduced against a large, already analyzed `.i64`, after a successful `analyze_function`, with auto-analysis and cache building disabled.

The local `decompile` implementation also performs a second `ida_hexrays.decompile()` call to collect refs after `decompile_function_safe()`, which may amplify the issue for expensive functions.

NeKroFR:
Closes #494.

A tool call longer than the health-probe timeout made the worker look dead to the supervisor, which dropped the session and killed the process mid-analysis.

**Why the obvious fix doesn't work?** idalib has no UI event loop, so `execute_sync` posted from a secondary thread is never dispatched. Just making the worker's HTTP server threaded deadlocks instead of fixing it.

**Worker:** the main thread now drains a job pump and HTTP moves to background threads. `@idasync` submits to the pump, so IDA still runs on the thread that owns the database while ping/health/cancel stay answerable. `server_health`
reports busy without touching IDA, the idle watchdog no longer counts an in-flight call as idle.

**Supervisor:** probes classify ok/busy/dead, busy is forwarded to, and only a worker silent past a grace window is reaped. Serving threaded keeps a busy worker from blocking calls to other sessions, which in turn required fixing a same-path concurrent `idb_open` race (the loser now gets the winner's session instead of a locked-database error).

**Side effect:** `notifications/cancelled` now works headless. Previously a busy worker couldn't receive it.

Tested on IDA 9.3

### DIFF
--- a/src/ida_pro_mcp/ida_mcp/api_core.py
+++ b/src/ida_pro_mcp/ida_mcp/api_core.py
@@ -20,6 +20,7 @@ import ida_nalt
 import ida_typeinf
 import idc
 
+from .mainthread import get_pump
 from .rpc import tool
 from .sync import idasync, get_tool_deadline
 from .utils import (
@@ -45,7 +46,7 @@ from .utils import (
 logger = logging.getLogger(__name__)
 
 
-class ServerHealthResult(TypedDict):
+class ServerHealthResult(TypedDict, total=False):
     status: str
     uptime_sec: float
     idb_path: str | None
@@ -56,6 +57,9 @@ class ServerHealthResult(TypedDict):
     hexrays_ready: bool
     strings_cache_ready: bool
     strings_cache_size: int
+    busy_tool: str
+    busy_sec: float
+    queued_calls: int
 
 
 class ServerWarmupStep(TypedDict, total=False):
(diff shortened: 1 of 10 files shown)


mrexodia:
Cheers!

PR #511, opened 2026-08-13
Stringify integers outside the JS safe-integer range

### LINKED ISSUE #510: 64-bit values sent as bare JSON numbers crash BigInt-aware MCP clients
## Summary

`tools/call` emits 64-bit integers as bare JSON numbers in `structuredContent`. Any value outside the JS safe-integer range (`|v| > 2**53 - 1`) becomes a `BigInt` in MCP clients that parse with a BigInt-preserving reviver, and the client then dies with `TypeError: Do not know how to serialize a BigInt` when it re-serializes the tool result.

In my case this killed the agent process mid-session, not just the one tool call.

## Reproduction

Any binary works — reading 8 bytes of an ASCII string as `u64` always exceeds `2**53 - 1`, so `tests/crackme03.elf` from this repo is enough. No arm64 target needed.

**1. Start the server:**

```bash
uv run idalib-mcp --host 127.0.0.1 --port 8745 tests/crackme03.elf
```

This serves streamable HTTP at `http://127.0.0.1:8745/mcp`. Since the supervisor requires `database=<session_id>` on every tool call, the repro first looks the session up via `idb_list`.

**2. Drive it with a Node-based MCP client:**

```bash
copilot -p "Use the ida MCP tools: first call idb_list to find the open session_id for crackme03.elf, then call get_int with that database, addr='0x318', ty='u64'. Report the exact returned value." --allow-all --additional-mcp-config '{"mcpServers":{"ida":{"type":"http","url":"http://127.0.0.1:8745/mcp","tools":["*"]}}}'
```

`0x318` is the first ≥8-byte string in `crackme03.elf`; the server returns:

```json
{"result": [{"addr": "0x318", "ty": "u64le", "value": 7795506888185572399}]}
```

`7795506888185572399 > 2**53 - 1`, so a BigInt-preserving parser turns it into `7795506888185572399n`, and the subsequent `JSON.stringify` throws `TypeError: Do not know how to serialize a BigInt`.

## Affected tools

Not specific to `get_int` — anything that can emit a raw 64-bit Python int:

- `get_int` with `ty: u64` / `i64` (easiest trigger, works on every binary)
- `extract_function_constants` in `utils.py` emits `"decimal": op.value` untyped; arm64 `o_imm` operands routinely exceed the safe range. This was my original failure, via `analyze_batch` with `include_constants: true`.
- enum member values in `api_types.py`

## Expected

`structuredContent` should never contain a JSON number outside `±(2**53 - 1)`.

## Prior art

[`bytebase/dbhub`](https://github.com/bytebase/dbhub) (Universal Database MCP Server) hit the same class of problem and solved it with a single choke point at the response-construction boundary:

- [`bigIntReplacer`](https://github.com/bytebase/dbhub/blob/ecf8c86f6e9fe1f3ba0b25b2ae08fa67c9e45f85/src/utils/response-formatter.ts#L6-L22) — values inside the safe range stay numbers, values outside become strings.
- Applied uniformly in [`createToolSuccessResponse`](https://github.com/bytebase/dbhub/blob/ecf8c86f6e9fe1f3ba0b25b2ae08fa67c9e45f85/src/utils/response-formatter.ts#L82-L92), [`createToolErrorResponse`](https://github.com/bytebase/dbhub/blob/ecf8c86f6e9fe1f3ba0b25b2ae08fa67c9e45f85/src/utils/response-formatter.ts#L66-L77), [`createResourceSuccessResponse`](https://github.com/bytebase/dbhub/blob/ecf8c86f6e9fe1f3ba0b25b2ae08fa67c9e45f85/src/utils/response-formatter.ts#L117-L131) and [`createResourceErrorResponse`](https://github.com/bytebase/dbhub/blob/ecf8c86f6e9fe1f3ba0b25b2ae08fa67c9e45f85/src/utils/response-formatter.ts#L97-L112).
- Pinned down by [`bigint-handling.test.ts`](https://github.com/bytebase/dbhub/blob/ecf8c86f6e9fe1f3ba0b25b2ae08fa67c9e45f85/src/utils/__tests__/bigint-handling.test.ts), including the boundary cases (`MAX_SAFE_INTEGER` stays a number, `MAX_SAFE_INTEGER + 1` becomes a string) and a mixed-row serialization test.
- Their MySQL connector enforces the same rule one layer down via [`supportBigNumbers`](https://github.com/bytebase/dbhub/blob/ecf8c86f6e9fe1f3ba0b25b2ae08fa67c9e45f85/src/connectors/mysql/index.ts#L66).

Their policy is value-driven (only values that actually don't fit get stringified), which is what I'm proposing here.

## Suggested fix

A single choke point in the `tools/call` middleware in `rpc.py`: recursively stringify ints outside the safe range (leaving `bool`, small ints, and existing hex strings alone), and rebuild the `content` text so it stays consistent with `structuredContent`.

Two output types then need widening to keep the sanitized value valid against the advertised `outputSchema`:

- `api_memory.IntReadResult.value`: `int | None` → `int | str | None`
- `api_types.EnumMemberUpsertResult.value`: `int` → `int | str`

Untyped `list[dict]` returns (e.g. the constants extractor) need no schema change.

## Environment

- ida-pro-mcp `main` @ 0b5f7ae
- IDA Pro 9.x via idalib, macOS
- Originally hit with GitHub Copilot CLI (Node client)

DzenIsRich:
Fixes #510.

## Problem

Tools returning 64-bit IDA values emit them as bare JSON numbers. MCP clients that parse responses with a BigInt-preserving reviver (needed to keep 64-bit address precision) get a `BigInt` back, then throw `TypeError: Do not know how to serialize a BigInt` on the next `JSON.stringify`. That kills the client process rather than failing a single tool call.

Reproduced on `tests/crackme03.elf`: `get_int` at `0x318` with `ty=u64` returns `7795506888185572399`, which is above `Number.MAX_SAFE_INTEGER`. Full repro steps are in #510.

Affected paths include u64/i64 reads (`api_memory.get_int`), arm64 `o_imm` immediates (`utils.extract_function_constants`), and enum values (`api_types`).

## Change

Sanitize once in the `tools/call` middleware in `rpc.py` — integers outside ±(2^53 - 1) go out as decimal strings. Both `structuredContent` and the `content` text are rewritten, since a client doing `JSON.parse(content[0].text)` with the same reviver hits the identical crash.

The fix is in `rpc.py` rather than `zeromcp/mcp.py`, where `structuredContent` is originally built, because `zeromcp` is vendored.

`IntReadResult.value` and `EnumMemberUpsertResult.value` are widened to `int | str` so sanitized output still validates against the advertised `outputSchema`.

The policy is value-driven, not type-driven: values within the safe range are untouched, so existing clients see no change for the overwhelming majority of responses.

## Prior art

`bytebase/dbhub` (Universal Database MCP Server) applies the same value-driven policy in [`src/utils/response-formatter.ts`](https://github.com/bytebase/dbhub/blob/ecf8c86f6e9fe1f3ba0b25b2ae08fa67c9e45f85/src/utils/response-formatter.ts#L6-L22), with [boundary tests](https://github.com/bytebase/dbhub/blob/ecf8c86f6e9fe1f3ba0b25b2ae08fa67c9e45f85/src/utils/__tests__/bigint-handling.test.ts).

## Tests

`tests/test_bigint_serialization.py` covers `sanitize_bigints` boundaries (including `bool` not being treated as `int`) and the middleware end-to-end, asserting the result still validates against the generated `outputSchema` and that `content[0].text` matches `structuredContent`.

`_install_tools_call_patch()` is renamed to `install_tools_call_middleware(server)` so tests can apply the shipped middleware to an isolated server instead of duplicating its body.

Full suite: 240 passed, 114 subtests passed.

## Notes for review

- `tests/test_mcp_spec_truncation.py` hand-reimplements this middleware and has drifted (it predates the sanitization step). Left alone here to keep the diff focused; happy to fold it into `install_tools_call_middleware` in this PR or a follow-up.
- The pure-Python suite cannot import `api_memory`, so the test file mirrors `IntReadResult` locally. That means it verifies the middleware and schema path but does not guard the real annotation against being reverted.


mrexodia:
This is a bug in the client.

DzenIsRich:
> This is a bug in the client.

I expected this reply. AFAICS most of the MCP servers try to workaround this.

mrexodia:
You can put a magic MCP proxy sanitizer between your broken client and ida-pro-mcp to work around Microsoft's terrible mess.

PR #513, opened 2026-08-15
fix(http): reply with HTTP/1.1 so pooled connections survive

### LINKED ISSUE #512: HTTP server replies with HTTP/1.0 and closes the socket, breaking clients that pool connections (Claude Code: ECONNRESET)
## Summary

`McpHttpRequestHandler` never sets `protocol_version`, so it inherits `BaseHTTPRequestHandler`'s default of `HTTP/1.0`. Python's `http.server` therefore closes the TCP connection after every response.

MCP clients that pool connections send `initialize`, then reuse the same socket for `notifications/initialized`. That socket is already closed server-side, so the second write fails. Claude Code reports:

```
ida: http://127.0.0.1:13337/mcp (HTTP) - × Failed to connect
  ECONNRESET: The socket connection was closed unexpectedly.
```

The server itself is healthy every individual request succeeds. Only connection reuse fails, so the failure looks like a connectivity problem rather than a framing one.

## Environment

- ida-pro-mcp 2.0.0 (`zeromcp/1.3.0`)
- IDA Pro 9.3.26.0327, Windows 11 Pro 26200
- Python 3.11.15
- Claude Code 2.1.220
- Plugin installed via `ida-pro-mcp --install` (GUI plugin, autostart, `127.0.0.1:13337`)

## Reproduction

Load any database so the plugin autostarts, then send two requests on one connection:

```powershell
$c = New-Object System.Net.Sockets.TcpClient("127.0.0.1", 13337)
$s = $c.GetStream()
$body = '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2025-06-18","capabilities":{},"clientInfo":{"name":"probe","version":"1"}}}'
$req = "POST /mcp HTTP/1.1`r`nHost: 127.0.0.1:13337`r`nAccept: application/json, text/event-stream`r`nContent-Type: application/json`r`nConnection: keep-alive`r`nContent-Length: $($body.Length)`r`n`r`n$body"
$b = [Text.Encoding]::ASCII.GetBytes($req); $s.Write($b,0,$b.Length); $s.Flush()
Start-Sleep -Milliseconds 900
$buf = New-Object byte[] 8192; $n = $s.Read($buf,0,8192)
(([Text.Encoding]::ASCII.GetString($buf,0,$n)) -split "`r`n")[0]   # HTTP/1.0 200 OK

# reuse the same socket
$body2 = '{"jsonrpc":"2.0","method":"notifications/initialized"}'
$req2 = "POST /mcp HTTP/1.1`r`nHost: 127.0.0.1:13337`r`nAccept: application/json, text/event-stream`r`nContent-Type: application/json`r`nContent-Length: $($body2.Length)`r`n`r`n$body2"
$b2 = [Text.Encoding]::ASCII.GetBytes($req2)
$s.Write($b2,0,$b2.Length); $s.Flush(); $s.Read($buf,0,8192)
```

**Expected:** second response returns `202 Accepted`.
**Actual:** first response is `HTTP/1.0 200 OK`; the second write fails with a connection-aborted error.

Note that `curl`, `Invoke-WebRequest`, and plain Node `fetch` all *pass* this test, because they open a fresh connection per request or correctly retire the HTTP/1.0 socket. This is why the bug appears client-specific.

## Wire capture

Proxying Claude Code through a logging TCP tap, on a single connection:

```
--- CLIENT->SERVER ---
POST /mcp HTTP/1.1
Accept: application/json, text/event-stream
User-Agent: claude-code/2.1.220 (claude-desktop, agent-sdk/0.3.229)
Connection: keep-alive
Host: 127.0.0.1:13399
Content-Length: 323

{"method":"initialize",...,"id":0}

--- SERVER->CLIENT ---
HTTP/1.0 200 OK
Server: zeromcp/1.3.0 Python/3.11.15
Content-Type: application/json
Content-Length: 234
Mcp-Session-Id: 8a8665d5-2d46-4d93-ac7b-11849136cb15

{"jsonrpc": "2.0", "result": {...}, "id": 0}

--- CLIENT->SERVER ---            <-- same connection
POST /mcp HTTP/1.1
mcp-session-id: 8a8665d5-2d46-4d93-ac7b-11849136cb15
Connection: keep-alive
Content-Length: 54

{"method":"notifications/initialized","jsonrpc":"2.0"}

===== UPSTREAM CLOSED =====       <-- server had already closed; client sees ECONNRESET
```

## Root cause

`McpHttpRequestHandler` (`src/ida_pro_mcp/ida_mcp/zeromcp/mcp.py`) does not set `protocol_version`, so it defaults to `"HTTP/1.0"` and `close_connection` is set after each response.

## Suggested fix

Opting into HTTP/1.1 requires every response to be self-delimiting. Most already send `Content-Length`; three paths do not.

1. `McpHttpRequestHandler`: add the class attribute

   ```python
   protocol_version = "HTTP/1.1"
   ```

2. The overridden `send_error` writes a body with no `Content-Length`:

   ```python
   def send_error(self, code, message=None, explain=None):
       body = f"{message}\n".encode("utf-8")
       self.send_response(code)
       self.send_header("Content-Type", "text/plain")
       self.send_header("Content-Length", str(len(body)))
       self.send_cors_headers()
       self.end_headers()
       self.wfile.write(body)
   ```

3. `do_OPTIONS` sends no body and no length add `self.send_header("Content-Length", "0")`.

4. `_handle_sse_get` has no `Content-Length` by design, so it must stay close-delimited. Change its `Connection: keep-alive` header to `Connection: close` (which also sets `close_connection` via `send_header`'s special handling).

Additionally, in `ida_mcp/http.py`, `_handle_config_post`'s `302` redirect sends no body and no `Content-Length`; add `Content-Length: 0`.

## Verification after patching

- Status line becomes `HTTP/1.1 200 OK`; the reuse repro above returns `202 Accepted` on the second request.
- `claude mcp list` reports `√ Connected`.
- `GET /sse` still returns `200 OK` with `Connection: close` and a valid `event: endpoint` frame.

## Workaround for other users

Until this is fixed, patching the installed package works (`ida-pro-mcp --install` symlinks the plugin, so edits in site-packages take effect on the next IDA start). Note that `uv tool upgrade ida-pro-mcp` reverts it.

xyOz-dev:
Fixes #512

## Problem

`McpHttpRequestHandler` never set `protocol_version`, so it inherited `BaseHTTPRequestHandler`'s `HTTP/1.0` default and Python's `http.server` closed the TCP connection after every response.

Clients that pool connections send `initialize`, then reuse the same socket for `notifications/initialized`. That socket is already closed server-side, so the second write fails. Claude Code reports `ECONNRESET` and the server appears unreachable, even though every individual request succeeds.

This is easy to miss because `curl`, `Invoke-WebRequest` and plain Node `fetch` all pass: they open a fresh connection per request or correctly retire the HTTP/1.0 socket.

## Changes

Opting into HTTP/1.1 requires every response to be self-delimiting, so four of the five changes are about framing rather than the version itself.

- `zeromcp/mcp.py`: set `protocol_version = "HTTP/1.1"` on `McpHttpRequestHandler`
- `zeromcp/mcp.py`: send `Content-Length` from the overridden `send_error`, which previously wrote a body with no length
- `zeromcp/mcp.py`: send `Content-Length: 0` from `do_OPTIONS`
- `zeromcp/mcp.py`: the SSE stream has no `Content-Length` by design, so it now sends `Connection: close` and stays close-delimited
- `ida_mcp/http.py`: send `Content-Length: 0` on the config page `302` redirect

Without the framing changes, flipping the version alone would replace the reset with a hang on those paths.

## Test

`test_streamable_http_reuses_pooled_connection` in `tests/test_server.py` asserts the response is HTTP/1.1, that the server does not signal close, and that a second request travels over the same socket. It compares the socket object because `http.client` silently reconnects, which would otherwise mask the bug.

Verified the test fails against the current behaviour and passes with the fix:

```
mode: PATCHED (protocol_version=HTTP/1.1)
  response.version=11 will_close=False
  second.status=202 sock_reused=True
RESULT: PASS

mode: UNPATCHED (protocol_version forced to HTTP/1.0)
  response.version=10 will_close=True
RESULT: FAIL -> Expected HTTP/1.1, got 10
```

## Manual verification

Against the GUI plugin on IDA 9.3 (Windows 11, Python 3.11.15):

- Status line is now `HTTP/1.1 200 OK`, and a second request on a reused socket returns `202 Accepted` instead of resetting
- `claude mcp list` reports the server as connected
- `GET /sse` still returns `200 OK` with `Connection: close` and a valid `event: endpoint` frame, so that transport is unaffected


mrexodia:
Added proper HTTP/1.1 support on main.

PR #517, opened 2026-08-29
fix: support Streamable HTTP session termination

psyche314:
## Motivation

Codex CLI sends `DELETE /mcp` with the `Mcp-Session-Id` header when it terminates a stateful Streamable HTTP session.

The current `zeromcp` HTTP handler does not implement `DELETE`, so Python's default handler returns HTTP 501. The MCP session has already worked successfully, but the client still receives an error during normal shutdown.

This behavior is covered by the MCP Streamable HTTP session-management specification:
https://modelcontextprotocol.io/specification/2025-06-18/basic/transports

## Changes

- Add `McpHttpRequestHandler.do_DELETE()` for `/mcp`.
- Validate `Mcp-Session-Id`.
- Remove the corresponding HTTP session from the server registry.
- Return HTTP 204 for successful termination.
- Return HTTP 400 when the session header is missing.
- Return HTTP 404 for an unknown session.
- Advertise `DELETE` in CORS preflight responses.
- Add an HTTP regression test for session termination.

This change does not alter JSON-RPC dispatch, session creation, SSE handling, or legacy HTTP+SSE behavior.

## Verification

- `python -m pytest -q tests/test_mcp_spec_http_e2e.py -k delete`
  - 1 passed
- Real Codex CLI connected to a local `psyche314` MCP server and a real `idat.exe` instance.
  - `instance_list`: passed
  - `server_health`: passed
  - IDA state: `ready`
  - Codex CLI shutdown completed without the previous `DELETE /mcp` HTTP 501 error

## Notes

Running the complete HTTP E2E file against the unmodified upstream baseline still exposes one unrelated pre-existing failure:

`test_single_threaded_stop_closes_idle_keepalive_connection`

That failure is outside the scope of this change.


mrexodia:
I would rather re-vendor the latest https://github.com/mrexodia/zeromcp version than fix this duplicate.

PR #518, opened 2026-08-29
Publish worker open state and session IDs in the discovery registry

jayluxferro:
## Summary

MCP clients that spawn a fresh `idalib-mcp` supervisor per session — or per tool call, as one-shot stdio gateways do — cannot see the workers a previous supervisor spawned: the supervisor's session map is in-memory only, and workers outlive it. The registry also goes dark *during* the open itself: a worker only registered after `open_binary` returned, and since auto-analysis pins the IDA main thread, a long analysis left no trace of which worker was opening what.

This PR makes workers self-describing in the existing discovery registry:

- workers register as `state: "opening"` with their pending `session_ids` **before** the blocking `open_binary`, refresh to `state: "ready"` with the full session list on success, and refresh/unregister on failure
- `register_instance()` gains optional `state`/`session_ids` parameters and preserves `started_at` across same-pid refreshes
- `InstanceInfo` documents both fields as optional; they are absent on GUI instances and older workers, and all readers use `.get()`, so the change is additive and backward-compatible

The supervisor-side consumer (session adoption across supervisor processes, `idb_list` enrichment, and an open budget for slow analysis) builds on this and will follow as a separate PR.

Related: #487 — orphaned idalib workers accumulate precisely because later supervisors cannot discover them; publishing worker identity and state in the registry is the foundation for reclaiming them.

## Test plan

- `uv run pytest tests/test_idalib_supervisor.py` — 65 passed on this branch (existing suite unchanged; the new fields are additive)
- Manual: opened a 93 MB binary under a short open budget — the registry entry showed `state: "opening"` with the session ID immediately, flipped to `ready` when analysis completed, and was removed on worker exit.


mrexodia:
I will redo the backend of ida-pro-mcp based on ida-nexus soon.

PR #522, opened 2026-09-12
fix(decompile): paginate large output within RPC limits

packmad:
Large decompilations could cross the 50k RPC limit, causing the visible code to be truncated while the cursor advanced past unseen lines. Some clients also do not surface the download URL stored in MCP metadata.

## Summary

- Add `max_lines`/`offset` pagination to `decompile`
- Keep pages below the RPC output limit
- Advance cursors only by lines actually returned
- Return refs on the first page and signal `refs_truncated`
- Enforce an aggregate 40k preview budget
- Preserve schema-safe download metadata


mrexodia:
the clients need to fix this

mrexodia:
Sorry I misunderstood, reopening!

Issue #497, opened 2026-07-28
GUI instance discovery file deleted moments after registration, even though instance is alive and reachable


vuhz:
## Summary

`discover_instances()` in `src/ida_pro_mcp/ida_mcp/discovery.py` deletes a GUI instance's `instance_<port>.json` registration file almost immediately after it's written — even though the instance is fully alive and reachable at the time. This breaks `idb_open(mode="prefer_gui"|"force_gui")` and `idb_list()` for adopting an already-running GUI session via `idalib-mcp`.

## Environment

- OS: Windows 11
- IDA Pro 9.3
- Installed via the Claude Code plugin marketplace (`claude plugin install ida-pro-mcp@mrexodia`), invoked as `uv run --project <plugin-cache> idalib-mcp --stdio`
- `idalib` activated via `py-activate-idalib.py` against the IDA 9.3 install

## Repro

1. Load the GUI plugin in IDA (`ida_mcp.py`, from this repo's `ida-plugin.json` entry point). Confirms in the IDA log:
   ```
   [MCP] Registered instance: ZDefend (pid=25252, port=13337)
     Discovery file: ...\Hex-Rays\IDA Pro\mcp\instances\instance_13337.json
   ```
2. Immediately (within a few seconds) call `idb_list` via `idalib-mcp --stdio`:
   ```json
   {"jsonrpc":"2.0","id":2,"method":"tools/call","params":{"name":"idb_list","arguments":{}}}
   ```
   Result: `{"sessions":[],"count":0}`
3. Check the instances directory — the file is gone:
   ```
   ls "$APPDATA/Hex-Rays/IDA Pro/mcp/instances/"
   # empty
   ```
4. At the same moment, the instance is still perfectly reachable:
   - `curl -X POST http://127.0.0.1:13337/mcp ...` → valid JSON-RPC response
   - Calling `server_health` directly against that port (via the non-idalib `ida-pro-mcp` stdio proxy, which doesn't go through this discovery/prune logic) returns real data (`module`, `idb_path`, `hexrays_ready: true`, etc.)

Reproduced this identically twice in one session (once against the original instance, once after re-triggering registration via Ctrl+Alt+M).

## What I ruled out

Directly isolated both liveness checks used by `discover_instances()` against the live PID/port right after a fresh registration, and both passed on their own:

```python
import discovery
discovery.is_pid_alive(25252)                       # -> True
discovery.probe_instance("127.0.0.1", 13337, 2.0)    # -> True
[...]

PR #525, opened 2026-09-20
fix(discovery): keep instances that bound a wildcard address

sxh313:
## Problem

`register_instance()` records the address the server **bound** to, and
`discover_instances()` later dials that address to decide whether the instance is
alive (`discovery.py:165`). A wildcard bind is perfectly valid to *listen* on but
not to *connect* to on Windows: Winsock fails such a connect immediately with
`WSAEADDRNOTAVAIL` instead of routing it to loopback the way BSD sockets do.

```
$ python -c "import socket; socket.create_connection(('0.0.0.0', PORT), timeout=2)"
OSError: [WinError 10049]            # WSAEADDRNOTAVAIL, raised in ~0.0s
$ python -c "import socket; socket.create_connection(('127.0.0.1', PORT), timeout=2)"
ok                                   # same listener, same process, still accepting
```
(Verified with a real listener accepting connections on `0.0.0.0`, so the failure is
about the address, not about load or timing. Windows 11, Python 3.12.13.)

Because the probe reports "unreachable", `discover_instances()` takes that as proof
of death and **deletes the registration file** of a live instance. Everything then
keeps working against `http://127.0.0.1:port` directly -- only the discovery path is
blind -- which matches the symptom shape of #497, but the cause here is the stored
address, not a transient probe miss: `is_pid_alive()` returns True and the connect
fails in 0.0s rather than after a timeout.

Reproduced end to end with the real functions (a live listener on an accepted port,
this process as the owner, registration written by `register_instance()`):

```
--- CASE 0 control: normal loopback instance
    bind=0.0.0.0:50394 backlog=5 accepting=True reg_host=127.0.0.1
    pid_alive=True discover_instances()=1 result(s) in 0.02s
    registration file still on disk: True
    VERDICT: kept

--- CASE 1 host recorded as 0.0.0.0
    bind=0.0.0.0:50396 backlog=5 accepting=True reg_host=0.0.0.0
    pid_alive=True discover_instances()=0 result(s) in 0.10s
    registration file still on disk: False
    VERDICT: PRUNED (live instance destroyed)
```

Non-loopback binds are a supported configuration here, not an accident: `serve()`
has dedicated `Host`-header handling for them (`zeromcp/mcp.py:201-213`), there is a
test file for it (`tests/test_browser_transport_guards.py` binds `0.0.0.0`), and
`#206` / `107c3f2` are about exactly this LAN use case. Anyone who reaches IDA from
another container by binding `0.0.0.0` on Windows loses their sessions from
`idb_list()` / session adoption, permanently for that process lifetime, since
nothing re-registers until the server restarts.

Two consumers read that same stored host and would fail against it even if the entry
survived: `idalib_supervisor.py:218` (`WorkerSession` reachability) and
`idalib_supervisor.py:1182`. Both take the value out of `discover_instances()`, so
fixing it there covers them too.

## Fix

Resolve the recorded bind address to a connectable one once, at read time:

```python
def _connectable_host(host: str) -> str:
    ...
    if address.is_unspecified:
        return "127.0.0.1" if address.version == 4 else "::1"
    return host
```

`discover_instances()` applies it right after the schema check, so the probe and
every consumer of the returned dict see `127.0.0.1` / `::1`. Addresses that are
already meaningful -- `127.0.0.1`, `::1`, a LAN IP like `192.168.1.10`, or a name
like `localhost` -- pass through untouched, so the normal configuration is
byte-for-byte unaffected, and this stays correct on macOS/Linux (where
`connect("0.0.0.0")` happens to work) rather than papering over a Windows quirk in
Windows-only code.

## Tests

New `tests/test_discovery_wildcard_host.py` (5 tests) covers the mapping plus the
pruning regression, and asserts the surviving entry carries a connectable host.

Against the previous code, the behavioural test fails for the reason described above:

```
$ PYTHONPATH=src python -m pytest tests/test_discovery_wildcard_host.py -q
FAILED ...::ConnectableHostTests::test_wildcard_binds_map_to_loopback
FAILED ...::ConnectableHostTests::test_other_hosts_pass_through
FAILED ...::DiscoverWildcardInstanceTests::test_live_instance_registered_on_wildcard_host_survives
  E  AssertionError: 0 != 1 : live instance was pruned as unreachable
3 failed, 2 passed in 0.24s
```

After the change:

```
$ PYTHONPATH=src python -m pytest tests -q
1 failed, 248 passed, 114 subtests passed in 20.48s
```

The single remaining failure is `test_mcp_spec_http_e2e.py::Http11TransportTests::
test_single_threaded_stop_closes_idle_keepalive_connection`. It is pre-existing on
this machine -- it fails identically on a clean checkout of `00e4aa7` before any edit
of mine -- and is unrelated to discovery: on Windows `shutdown(SHUT_RDWR)` does not
wake a thread already blocked in a `makefile()`-backed read, so `stop()` waits for
the client. I'll file that separately rather than mix it into this one.

## How to verify

1. `git checkout fix/discovery-wildcard-host && PYTHONPATH=src python -m pytest tests/test_discovery_wildcard_host.py -q` -> 5 passed.
2. `git stash push -- src/ida_pro_mcp/ida_mcp/discovery.py` and rerun -> 3 failed, the third with `0 != 1 : live instance was pruned as unreachable`.
3. On Windows, run any MCP server here with host `0.0.0.0` (e.g. `--host 0.0.0.0`),
   then call `discover_instances()` from another process: before the change the
   instance is missing and its `instance_<port>.json` is gone from
   `%APPDATA%\Hex-Rays\IDA Pro\mcp\instances\`.


mrexodia:
Why are you binding to `0.0.0.0`, what is the use case?

sxh313:
Short answer: I don't - nothing in this repo binds to a wildcard by default. `idalib_server.py` takes `--host` with `default="127.0.0.1"` (src/ida_pro_mcp/idalib_server.py:194-196), and the supervisor always spawns its own workers with a literal `--host 127.0.0.1` (src/ida_pro_mcp/idalib_supervisor.py:253-259), so a `0.0.0.0` / `::` in an instance file only ever appears because a user ran the server that way deliberately.

That is the supported LAN case the project already reasons about: `ida_mcp/http.py:196-215` documents that "non-loopback servers (`0.0.0.0`) accept any Host" and that `unrestricted` CORS is what "browser-based LAN config access" needs - i.e. exposing the MCP endpoint to another machine, a container, or WSL2 while IDA runs on Windows. When someone does that, `main()` stores `args.host` in `_BOUND_HOST` (:234-235) and hands it straight to `register_instance()` (:248, and the deferred registration at :147-148), so the registration file records the **bind** address.

The defect is that a bind address is not necessarily a connect address. Winsock rejects `connect(("0.0.0.0", port))` and `connect(("::", port))` with `WSAEADDRNOTAVAIL` (10049) - measured on this Windows 11 box against a listener that `127.0.0.1` connects to fine - whereas BSD sockets route the wildcard to loopback, so the same file works by accident on Linux/macOS. On Windows `discover_instances()` fails its own secondary probe and then `os.unlink()`s the registration file of a live instance, destroying the idb path and pid that the supervisor's adoption path needs.

Why normalize on the read side instead of at write time:
- The consumers of a discovered instance treat `host` as a connect target (`probe_instance(inst["host"], ...)` in `idalib_supervisor.py`, and `_make_gui_session` / `_adopt_worker_instance` build sessions from it), so the read path is exactly where a bind address has to become a connect address for every caller at once.
- Registering `127.0.0.1` while actually bound to `0.0.0.0` would throw away information the operator chose (that the instance is reachable off-host), and would not help the files users already have on disk.
- Special-casing only `probe_instance()` would keep the file but still hand every consumer an unconnectable host.

The mapping is in-memory only (`info["host"]`), so the file on disk still records the real bind address. If you would rather see the wildcard rejected at `register_instance()` with an explicit `--advertise-host`, say so and I will rework it - the test in this PR pins the observable behavior either way.

mrexodia:
I don't understand, what is the point of a PR if you do not have a use case for the fix yourself?

sxh313:
You're right to call it out: I don't have one. My setup is loopback (`127.0.0.1`, which is the default on every entry point), I have no IDA install behind an off-host bind, and the repro in the test is synthetic. I found this by reading `discovery.py` against Winsock semantics, not by hitting it.

The person who *would* have the use case is reachable through code you already ship: `ida_mcp/http.py:196-215` treats a `0.0.0.0` bind as a supported configuration ("non-loopback servers accept any Host", `unrestricted` CORS for "browser-based LAN config access"), `tests/test_browser_transport_guards.py:147` unit-tests `_host_header_allowed_for_bind("0.0.0.0", ...)` under a "LAN connectivity" heading, and the GUI config form lets you type that host. So the bind is anticipated in the security layer but documented nowhere in the README, and nobody has reported the consequence on Windows - that the registration file of a live instance gets `os.unlink()`ed.

So by your standard this is speculative hardening, not a bug with a reporter behind it, and I'm not going to argue it out of your queue. Close it if you'd rather not spend review time on it; I'd make the same call. Either way I won't send more socket-portability cleanups of this shape unless someone actually hits one.

The one thing I'd keep regardless of this PR is the asymmetry itself: `connect("0.0.0.0")` works on Linux/macOS and fails with `WSAEADDRNOTAVAIL` on Windows, so any wildcard-bound instance is silently broken on one platform only. If you'd rather see that raised as an issue for whoever deploys that way, say so and I'll file it instead of opening a PR.

